PR #2, opened 2023-09-13
Set up basic GH actions workflow for linting and Jest tests

marktlinn:
• Add workflow 
• Jest example test 
• Update Jest path regex


alcb1310:
Merging this PR will close this issue
https://suwanna.atlassian.net/browse/CGD-50

.github/test.yml
```diff
@@ -0,0 +1,27 @@
+name: Lint & Test Workflow
+
+on:
+  pull_request:
+    branches: [main, dev]
+
+jobs:
+  build:
```
alcb1310, line 8: between lines 7 and 8 there should be a job name that will group all of the steps under eg:
jobs:
    

```yaml
jobs:
    run_tests:
        build:
           <the rest>
```

alcb1310 (changes requested):
Haven't found a way to be able to run this action or to add it the required steps
Also, I believe the `test.yml`file should be under `.github/workflows` folder

marktlinn:
> Haven't found a way to be able to run this action or to add it the required steps Also, I believe the `test.yml`file should be under `.github/workflows` folder

Thanks for the review, Andresc. That's my bad it didn't start running automatically, I dragged across the file, it should've been in the worflows folder is all.

.github/test.yml
marktlinn: I see what you mean, for this workflow though we have the name of the overall workflow as a title (at the top of the file) and then each step has a name. I think as it is, that works ok, as they're all tied together by the title name. We could always add names for jobs too, if we create more jobs at a later date, but for now I think it's ok without. 

What do you think?

alcb1310: I like the idea of having one test per file, however if we have many workflows it will spin up a new server for each file, and that will take some time, for now I believe that is fine but we'll see at a later date if we require to change the yaml structure to have the name of the test inside the jobs property

alcb1310: @marktlinn I will mark this as resolved however, I'd rather have the person who creates the PR to be the one who merges it, so I will leave that to you

alcb1310 (approved):
All required changes were addressed so I approve this PR

Dan-Y-Ko:
Will the deployment actions be handled in a separate pr?

PR #6, opened 2023-09-18
Docker Dev and Test containers + integration test config + scripts

marktlinn:
• Setup integration tests
• Created two Docker images, one to handle dev and the other tests (also added scripts to automate starting containers, running integration tests then tearing down containers - we can later add e2e tests to this if wanted?)
• Created a simple endpoint for quick heathchecks through curl to the "/health-check".
• Updated readme to reflect changes + Docker configs.

*I've added the .env.test and .env.dev variables into the readme, as they were in the readme in the FE repo anyway, and they're only good for local development. Many are hardcoded into the Docker images anyway, but have set them just incase that changes at some point down the line.*

### DIFF
--- a/README.md.old
+++ b/README.md.old
@@ -1,2 +1,0 @@
-# chingu-dashboard-be
-Chingu Dashboard Moonshot (backend)


README.md
```diff
@@ -65,6 +65,114 @@ $ yarn run test
...
+```
+
+This command will spin up the test Postgres container, run your tests and tear down the containers for you in one command.
+
+#### <a name="envfiles">Dev and Testing environment variable files</a>
+These files are only for development and testing purposes, set them in your root directory. They're pretty much the same and many of the variables are actually hardcoded into the Docker files for convenience, but that could change in future so be sure to copy across the entire files.
+
+```bash
+# .env.test
+DATABASE_URL=postgresql://chingu:chingu@localhost:5433/dashboard?schema=public
```
alcb1310, line 151: Both development and test database URL are the same, shouldn't we have a different database URL for the test environment? that way we can cleanup the database so we can have a clean state for the tests, however if this is just an example of the environment variables, it should explicit say so and also that a different db should be used in the testing enviornment

docker-compose-test.yml
```diff
@@ -0,0 +1,23 @@
...
+  test-pgdb:
+    container_name: test-pgdb
+    image: postgres:15-alpine
+    ports:
+      - "5433:5433"
+    hostname: localhost
+    environment:
+      POSTGRES_USER: chingu
+      POSTGRES_PASSWORD: chingu
+      POSTGRES_DB: dashboard
```
alcb1310, line 12: Should we have a different db name here than in the dev environment?

src/prisma/prisma.service.ts
```diff
@@ -1,13 +1,14 @@
 import { Injectable } from "@nestjs/common";
 import { PrismaClient } from "@prisma/client";
 
+console.log("HEHREHREHR: => ", process.env.DATABASE_URL);
```
alcb1310, line 4: Please remove this console log

alcb1310 (changes requested):
Leaved comments on the files but in summary:
 - in src/prisma/prisma.service.ts => remove the console.log
 - You have the same database name for both test an development, shouldn't those be different?

README.md
marktlinn: They should be different, one is `@localhost` & the other is `@chingu`, which was the original URL used in the FE Docker file and now used for Dev here. The localhost being for tests

src/prisma/prisma.service.ts
marktlinn: Oops 🤪, of course!

docker-compose-test.yml
marktlinn: I think it should be alright as is, as they'll be running on different container names with different `DATABASE_URL`s and use the same ports so couldn't run contemporaneously, anyway.

Dan-Y-Ko:
> > Only thing I'm confused about here is you have hardcoded db config values but also suggest to add a env file for test and dev. Is there a benefit to hardcoding the db config into the docker compose file?
> 
> I don't think it makes a much difference, as the values for the .env would be in the readme anyway and they'll only be used for dev and testing, so there's no prod environment variables on display. I just copied across the .envs in full in case the config ever changes, but could just change it to only having the DATABASE_URL values in there. Whatever the people prefer.

oh ok. Yea I don't see us using docker for prod so I don't think we need to use env in there. As for the env, I see what you did. I guess we can leave it in the readme for now since it won't particularly do anything but will be necessary if we ever do decide to use it.

Dan-Y-Ko (commented):
I will test this tomorrow. Can you resolve the conflicts in the meantime.

Dan-Y-Ko (commented):
Couple issues.

1. the yarn studio command is hard coded to test environment. Is that intended?
2. yarn migrate:test gives me an error.  https://i.imgur.com/SwgkTQW.png
3. DB gets wiped when we shut down compose. I'd rather we keep the volumes for dev environment

Dan-Y-Ko:
> > 2. https://i.imgur.com/SwgkTQW.png
> 
> Hi Dan, cheers for this.
> 
> 1. I didn’t set up the studio script, but I presume that was intentional. It’s only really changing the target .env file so can just create another script `studio:test`, `studio:dev` which target separate files.
> 2. Is this error occurring after running the `start:docker:test:all`? With Docker running then running `yarn migrate:test`? If so, could you check your firewall to ensure it’s allowing connections to/from Docker?
> 3. Were you using the ‘down’ script in the package JSON? That should tear down the docker containers completely, if you use the `docker compose -f <file-name> down` it should persist the volumes. Can swap that out to be the default in the scripts if that’s helpful?

1. Could you add a script to run prisma studio for dev enviroment and add that to readme instructions, or just remove it and update the readme to reflect that as well? I was following the readme and didn't notice that the studio command was hardcoded to the test environment.
2. Yes, it is not being blocked by firewall. Am I supposed to run the yarn migrate:test command after doing yarn docker:start:dev? cuz it works there.
3. Yes, can you make it default so it doesn't tear it down when using down command, for dev.

The goal of these changes is just to make it easier for anyone new coming to our project and may not be super knowledgeable about backend and / or docker so they can just have the scripts as an easy safety net to get started fast.

PR #7, opened 2023-09-23
Feature/team directory endpoints

cherylli:
Made some changes to the database
created a few endpoints, which can be seen/tested from the swagger docs
I've also added global error handling
I've pulled Dan's branch into mine so it has some of his commits. So ignore his endpoint (ideation) for this PR

image

teams -> get all teams
teams/voyage/:id -> get all teams in that voyage
teams/:id -> get the team info by team Id
teams/:id/members -> get all the members of the team
PATCH teams/:teamId/member/:userId -> update hrPerSprint

first 2 is probably not needed for frontend but will be needed for admin purpose so i'll just leave them

image

image
image


cherylli:
might need to disable jest test till we put tests in. The nest generated tests are failing

prisma/migrations/20230920002613_just_trying_to_get_the_sql/migration.sql
Dan-Y-Ko, line 1: This should be deleted.

src/prisma-client-exception/prisma-client-exception.filter.ts
Dan-Y-Ko, line 1: Is this going to catch any exceptions at the server level? This looks to me like a prisma / db specific exception handler.

Dan-Y-Ko (changes requested):
I left some initial thought so far. Will test again later but other than my comments so far, it looks good.

src/prisma-client-exception/prisma-client-exception.filter.ts
cherylli: no that's just prisma, we can add another filter for other exceptions, we will also need to add cases for other prisma errors in future. Otherwise, any error would just be 500 internal server error by default

Dan-Y-Ko:
Only question I have, would it not be better to have a server level exception filter? Something like this: https://docs.nestjs.com/exception-filters#catch-everything. Any error from prisma should be passed up to this filter as well.

cherylli:
> Good work with this. How are the tests running for you locally, is everything passing?
> 
> going forward, I'd be careful merging other active branches into your current working branch as it can cause issues in the future which can be difficult to debug, it also can cause the PR to grow in size and complexity as well. If there's a feature that's a blocker for you in a branch you're going to work on, it's better to have that feature taken out and worked on as a smaller ticket so that it can be completed, reviewed and merged. If it isn't a blocker it can be merged independently and then the local branch you're working on can be rebased to be kept up-to-date with the primary branch (`dev` in our case).

probably not a good idea in general but I didn't think it was a huge issue this time as we were working on different branches but there were some changes needed which would be the same in my PR as well (swagger setup for example).

cherylli:
> Only question I have, would it not be better to have a server level exception filter? Something like this: https://docs.nestjs.com/exception-filters#catch-everything. Any error from prisma should be passed up to this filter as well.

I think you can have both, this server level exception would catch anything not catched by the prisma specific filter. I think it's better to have a prisma specific one as well to have seperate error handling for different things instead of putting everything in a single error handling module, especially when the app grows

cherylli:
> I'd say if we have something that needs to be shared e.g. swagger. Best thing to do would be just do a micro PR with swagger and merge it. It's not a big deal here but just as a best practice it's generally better to avoid is all I was saying.

yeah that's probably what i'd do, provided it doesn't take days/weeks to merge that "micro PR"

Dan-Y-Ko:
> Was able to test all the endpoints with the new Docker fixes and everything worked well, was able to get user data, update team member's hourly commitment to sprints which persisted in the DB.
> 
> Just need to run a `yarn format` to run prettier and remove any style errors/warnings. After that I think it's ready to merge 👍. Nice work!

Is there husky not working for that?

cherylli:
> Was able to test all the endpoints with the new Docker fixes and everything worked well, was able to get user data, update team member's hourly commitment to sprints which persisted in the DB.
> 
> Just need to run a `yarn format` to run prettier and remove any style errors/warnings. After that I think it's ready to merge 👍. Nice work!

still not working for me, but I don't see any changes in `docker-compose-dev/yml` in the last PR, only `docker-compose-test.yml` is that what we're supposed to run?

cherylli:
Ok maybe I was missing something, will try again later but if it works for you then we can probably merge this

jenny-alexander:
[CGD-49 Team Directory Endpoint](https://trello.com/c/mZzKxh6T/3-cgd-49-team-directory-endpoint)

Issue #18, opened 2023-10-16
Create resources endpoints


jenny-alexander:
Requirements can be found here: https://suwanna.atlassian.net/wiki/spaces/SD/pages/13828097/Features+Page

jenny-alexander:
Requirements can be found here: https://suwanna.atlassian.net/wiki/spaces/SD/pages/13828097/Features+Page

jenny-alexander:
![](https://github.trello.services/images/mini-trello-icon.png) [CGD-18 Create Resources endpoints](https://trello.com/c/yjiIglUr/22-cgd-18-create-resources-endpoints)

cherylli:
closing as PR merged

PR #27, opened 2023-10-25
Features/team resources endpoints

curtwl:
# Description

Create endpoints for team resources

## Issue link

Fixes # [18](https://github.com/chingu-x/chingu-dashboard-be/issues/18)

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Tested with Swagger and Postman. 

# Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] Any dependent changes have been merged and published in downstream modules


src/resources/resources.service.ts
```diff
@@ -0,0 +1,98 @@
+import { BadRequestException, Injectable, UnauthorizedException } from '@nestjs/common';
+import { CreateResourceDto } from './dto/create-resource.dto';
+import { UpdateResourceDto } from './dto/update-resource.dto';
+import { PrismaService } from 'src/prisma/prisma.service';
+import { TeamResource } from '@prisma/client';
+
+const userId = 6;
```
cherylli, line 7: how come userId is hardcoded? for now, it can come from request body till we have auth sort which will pass the userId

I believe this userId should also be a UUID, not int

there's a difference between userId, and voyageTeamMemberId. UserId is like a Chingu member user id, while voyageTeamMemberId is their IDs in voyages. It is set up this way because a user can participate in multiple voyages

cherylli (dismissed):
you can run `yarn format` to fix lint errors

src/resources/resources.service.ts
cherylli: other reasons you can't hardcode here is the endpoint will basically be usable by frontend, and Ids can change everytime the data is reseeded

curtwl: I guess just a bad habit from using workarounds like that before while solo coding, it makes complete sense to me now why that's not workable.

Also completely clear to me now the difference between the voyageTeamMemberId and userId.

jenny-alexander (approved):
LGTM - great job taking in all comments and incorporating them into your code.
I tested endpoints locally.
_Endpoints should be plural - we will take care of that in another task_

src/resources/resources.controller.ts
jenny-alexander, line 1: Can you change the URI naming from teamResources to team-resources?

src/resources/dto/delete-resource.dto.ts
jenny-alexander, line 1: As noted by Cheryl and Dan in discord, we'll remove the userId from the delete request once auth is working.

src/resources/dto/create-resource.dto.ts
jenny-alexander, line 1: Are url and title allowed to be empty? I see in the db that they are defined with NOT NULL constraint. Can you add a decorator from class-validator checking if they are empty?
`import { IsNotEmpty } from "class-validator";`

src/resources/resources.service.ts
jenny-alexander, line 1: Should you return something if delete wasn't successful? A descriptive message for FE?

curtwl: I added in a NotFoundException for delete and update if the resource wasn't found, in the same function that checks if the user is authorized to perform the action. I also added a try/catch block for delete that throws a generic BadRequestException if anything else happens, not sure if that's appropriate.

jenny-alexander (approved):
Reviewed changes - looks good.
I tested 2 more random API endpoints within team resources to make sure they work as expected.

Only outstanding is the userId within body of resource DTO - will this be removed once auth is in? Shouldn't be in there as it doesn't get modified on update.

Dan-Y-Ko:
> Reviewed changes - looks good. I tested 2 more random API endpoints within team resources to make sure they work as expected.
> 
> Only outstanding is the userId within body of resource DTO - will this be removed once auth is in?

That's correct.

I haven't looked into this in detail but I think it'd be good to name it voyage resources or something like that since we have this resources and the global resources. Idk what naming convention be team decided on but it looks like there's another folder using kebab casing, so voyage-resources? Something like that.

cherylli:
> > Reviewed changes - looks good. I tested 2 more random API endpoints within team resources to make sure they work as expected.
> > Only outstanding is the userId within body of resource DTO - will this be removed once auth is in?
> 
> That's correct.
> 
> I haven't looked into this in detail but I think it'd be good to name the folder voyage resources or something like that since we have this resources and the global resources. Idk what naming convention be team decided on but it looks like there's another folder using kebab casing, so voyage-resources? Something like that.

yes nestjs use kebab case for files and folders

src/resources/resources.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+@Controller("team-resources")
+@ApiTags("team resources")
+export class ResourcesController {
+    constructor(private readonly resourcesService: ResourcesService) {}
+
+    @ApiOperation({
+        description:
+            "Adds a URL with title to the team's resources, addedBy: teamMemberId (int)",
+    })
+    @Post(":teamMemberId")
```
cherylli, line 26: How likely is it for frontend to store teamMemberId? will it be stored in localstorage etc? Another concern is when people participate in multiple teams, they would have multiple `teamMemberId` so (teamId + userId) would be the best way to gaurantee uniqueness from the frontend (User Id would be in the auth header, TeamId would probably already be in the url params, both of these are readily available). We have been using userId + teamId in other endpoints (since teamId would be somewhere in the frontend because a lot of other things use it)

src/resources/resources.controller.ts
```diff
@@ -0,0 +1,74 @@
...
+    @Get(":teamId")
+    findAllResources(@Param("teamId", ParseIntPipe) teamId: number) {
+        return this.resourcesService.findAllResources(teamId);
+    }
+
+    @ApiOperation({
+        description:
+            "Edit URL/title for a resource if teamMemberId (int) matches logged in user",
+    })
+    @Patch(":teamMemberId/:resourceId")
```
cherylli, line 49: this endpoint probably doesn't follow the naming convention we have been following
https://restfulapi.net/resource-naming/

cherylli (commented):
I see you are solely using teamMemberId, in this case, userId won't be required at all. But I've explained why I believe using userId + team Id to find teamMemberId in the backend would be better in another comment

src/resources/resources.controller.ts
curtwl: Should I change this to @Post(":teamId/:userId") then? Couldn't find an example like this searching the codebase.

src/resources/resources.controller.ts
curtwl: I don't see which convention is being broken -- seems like the kebab case convention doesn't apply to param names?

src/resources/resources.controller.ts
cherylli: would be teams/:teamId. userId would come from the auth header, but for now it can be in the body
both ideation and tech stack endpoints have something similar to this

src/resources/resources.controller.ts
cherylli: you can read the linked article for example, and the rest of the project we have been using this, and Kenny has fixed some older ones in the recent PR #28 

it would be something like `team-members/:teamMemberId/resources/:resourceId` in this case. but as discussed we should be passing teamId instead of teamMemberId here

curtwl: Updated the code following the way ideations did it:
/api/v1/teams/{teamId}/team-resources    for POST/GET
/api/v1/team-resources/{resourceId}    for PATCH/DELETE

Wasn't sure if perhaps PATCH/DELETE should be /api/v1/teams/{teamId}/team-resources/{resourceId} instead, ideations did it the like the first way and tech stacks the second way.

jenny-alexander (dismissed):
LGTM

jenny-alexander (approved):
Changed approved

smurph7894:
Tested each of the endpoints and they are working an protect for creator only edits when required.

smurph7894:
Do we want to consider having a validator that a url is bring input? It looks like prisma does work with regex and other options. https://www.prisma.io/docs/concepts/components/prisma-client/custom-validation

cherylli:
> Do we want to consider having a validator that a url is bring input? It looks like prisma does work with regex and other options. https://www.prisma.io/docs/concepts/components/prisma-client/custom-validation

nestjs has a @isUrl

cherylli:
> > Do we want to consider having a validator that a url is bring input? It looks like prisma does work with regex and other options. https://www.prisma.io/docs/concepts/components/prisma-client/custom-validation
> 
> nestjs has a `@IsUrl`

cherylli:
Looks good but missing some basic error handling for the POST and GET, like if TeamId doesn't exist it just returns a generic 500 error. 

I also agree to adding something to validate it's a url, "IsUrl"

and for GET, if the team doesn't exist it returns [] which should probably return an error, as opposed to returning [] when there's actually no votes

kennytram (approved):
lgtm

PR #33, opened 2023-11-01
update response json in add vote endpoint

Dan-Y-Ko:
# Description

Added member object as part of response in add vote endpoint. This data is needed to display the member's avatar in the voteds card in the UI.

An alternative approach that I know of, is to fetch the data using the get endpoint but from what I know, this is not recommended because we'd have to make another http call to the server and on top of that, we'd be getting extra data we don't need. It's better to make a db call to get the specific data we need and return that as additional data in the post request.

This will probably need to be done in the create ideation endpoint as well but can figure that out when I get to that part in the FE.

This is a urgent issue and is blocking me atm, so needs to be resolved asap.

## Issue link

Fixes # (issue)

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Tested with swagger and also with the FE. Everything is working as intended.

# Checklist:

- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/ideations/ideations.service.ts
+++ b/src/ideations/ideations.service.ts
@@ -76,8 +76,6 @@ export class IdeationsService {
                     member: {
                         select: {
                             id: true,
-                            firstName: true,
-                            lastName: true,
                             avatar: true,
                         },
                     },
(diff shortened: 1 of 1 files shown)


jenny-alexander:
I set the PR to draft because it's on hold for now.

jenny-alexander:
Closing this as it isn't needed anymore. If something like this is needed in the future, FE team will reach out to us.

PR #60, opened 2023-11-20
Merge Dev branch into Main

jenny-alexander:
# Description

Merge Development branch into Main branch


cherylli:
Probably need to wait till we merge Kenny's test PRs
image

this should be fixed in those PRs - e2e can't read absolute paths

cherylli:
do we want to run the e2e tests on main as well?
looks like this only runs the unit tests

kennytram:
> do we want to run the e2e tests on main as well? looks like this only runs the unit tests

We will need to set it up to work with the Test DB first before deciding to run e2e tests on the main.  I don't have access to this project's settings, so I wouldn't know if it is already configured or not.

cherylli:
> > do we want to run the e2e tests on main as well? looks like this only runs the unit tests
> 
> We will need to set it up to work with the Test DB first before deciding to run e2e tests on main. I don't have access to this project's settings, so I wouldn't know if it is already configured or not.

what do you need? is it just the test `DATABASE_URL` github repo secret? If so, Jenny can set that up

cherylli (approved):

PR #73, opened 2023-12-15
Feature/features order

### LINKED ISSUE #64: Add 'order' field to project features
A Chingu member needs to have the ability to order features within each 'bucket' (Must Have, Should Have, Nice to Have).

See Trello card for more details.

smurph7894:
# Description

Updated Feature Module to include ordering per team and category.  Note all but one service has changes to take into account their order. UpdateFeatureOrderAndCategory is open to team members to manipulate per FE request. 

## Issue link

Fixes #64 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

Tested in swagger.

# Checklist:

### DIFF
--- a/src/features/dto/update-feature-order-and-category.dto.ts
+++ b/src/features/dto/update-feature-order-and-category.dto.ts
@@ -1,12 +1,14 @@
-import { IsNotEmpty, IsOptional } from "class-validator";
+import { IsNotEmpty, IsNumber, IsOptional } from "class-validator";
 import { ApiProperty, ApiPropertyOptional } from "@nestjs/swagger";
 
 export class UpdateFeatureOrderAndCategoryDto {
     @IsNotEmpty()
     @ApiProperty()
+    @IsNumber()
     order: number;
 
     @IsOptional()
     @ApiPropertyOptional()
+    @IsNumber()
     featureCategoryId: number;
 }
--- a/src/features/dto/update-feature-order-and-category.dto.ts
+++ b/src/features/dto/update-feature-order-and-category.dto.ts
@@ -1,8 +1,11 @@
 import { IsNotEmpty } from "class-validator";
 import { ApiProperty } from "@nestjs/swagger";
 
-export class UpdateFeatureOrderDto {
+export class UpdateFeatureOrderAndCategoryDto {
     @IsNotEmpty()
     @ApiProperty()
     order: number;
+
+    @ApiProperty()
+    featureCategoryId: number;
 }
--- a/src/features/dto/update-feature-order-and-category.dto.ts
+++ b/src/features/dto/update-feature-order-and-category.dto.ts
@@ -1,11 +1,12 @@
-import { IsNotEmpty } from "class-validator";
-import { ApiProperty } from "@nestjs/swagger";
+import { IsNotEmpty, IsOptional } from "class-validator";
+import { ApiProperty, ApiPropertyOptional } from "@nestjs/swagger";
 
 export class UpdateFeatureOrderAndCategoryDto {
     @IsNotEmpty()
     @ApiProperty()
     order: number;
 
-    @ApiProperty()
+    @IsOptional()
+    @ApiPropertyOptional()
     featureCategoryId: number;
 }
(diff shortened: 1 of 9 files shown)


cherylli (commented):
with the new changes we would have a PATCH endpoint for updating "description" and another endpoint for updating the order, am I correct in understanding this? 

Would it be better to have 1 endpoint where you can change both? depends on what is in the request body

smurph7894:
> with the new changes we would have a PATCH endpoint for updating "description" and another endpoint for updating the order, am I correct in understanding this?
> 
> Would it be better to have 1 endpoint where you can change both? depends on what is in the request body

I talked to  Jane to get requirements and there are 2 separate updates needed. Our normal one for the creator to update the description and another for all team members to be able to drag and drop features to different orders and categories.

cherylli:
> > with the new changes we would have a PATCH endpoint for updating "description" and another endpoint for updating the order, am I correct in understanding this?
> > Would it be better to have 1 endpoint where you can change both? depends on what is in the request body
> 
> I talked to Jane to get requirements and there are 2 separate updates needed. Our normal one for the creator to update the description and another for all team members to be able to drag and drop features to different orders and categories.

you would still be able to do both in the same endpoint. If we have to add more things to update, would we make a new endpoint for each thing we need to update? What does everyone else think?

smurph7894:
> > > with the new changes we would have a PATCH endpoint for updating "description" and another endpoint for updating the order, am I correct in understanding this?
> > > Would it be better to have 1 endpoint where you can change both? depends on what is in the request body
> > 
> > 
> > I talked to Jane to get requirements and there are 2 separate updates needed. Our normal one for the creator to update the description and another for all team members to be able to drag and drop features to different orders and categories.
> 
> you would still be able to do both in the same endpoint. If we have to add more things to update, would we make a new endpoint for each thing we need to update? What does everyone else think?

It would depend on what the update is and what makes sense for the situation if you add another endpoint or not. 

I have them currently as two separate as it made sense since the UI actions are separate. Also the order and category update is more complex and both would be easier to maintain if kept to single responsibility endpoints. I felt more conditionals being added to combine them would add more complexity and room for error. Is there a benefit to combining?

cherylli:
> > > > with the new changes we would have a PATCH endpoint for updating "description" and another endpoint for updating the order, am I correct in understanding this?
> > > > Would it be better to have 1 endpoint where you can change both? depends on what is in the request body
> > > 
> > > 
> > > I talked to Jane to get requirements and there are 2 separate updates needed. Our normal one for the creator to update the description and another for all team members to be able to drag and drop features to different orders and categories.
> > 
> > 
> > you would still be able to do both in the same endpoint. If we have to add more things to update, would we make a new endpoint for each thing we need to update? What does everyone else think?
> 
> It would depend on what the update is and what makes sense for the situation if you add another endpoint or not.
> 
> I have them currently as two separate as it made sense since the UI actions are separate. Also the order and category update is more complex and both would be easier to maintain if kept to single responsibility endpoints. I felt more conditionals being added to combine them would add more complexity and room for error. Is there a benefit to combining?

no, just feel cleaning to have one endpoint doing all the updating. At first glance, /reorder feels like it should just do re-ordering but it also updates the featureCategory. 

or I felt it was not clear why there are 2 endpoints to update different things

src/features/dto/update-feature-order-and-category.dto.ts
```diff
@@ -0,0 +1,12 @@
+import { IsNotEmpty, IsOptional } from "class-validator";
+import { ApiProperty, ApiPropertyOptional } from "@nestjs/swagger";
+
+export class UpdateFeatureOrderAndCategoryDto {
+    @IsNotEmpty()
+    @ApiProperty()
+    order: number;
```
cherylli, line 7: I think we need to add @IsNumber() here

cherylli (approved):
Works fine but no error message when an invalid categoryId is entered

cherylli:
[CGD-64: Add "order" field to project features](https://trello.com/c/mQeNfmbO/71-cgd-64-add-order-field-to-project-features)

kennytram:
> endpoint for each thing we need to update? What does everyone else think?

The first PATCH endpoint only lets the creator to update the feature, whereas this PATCH endpoint will let anyone on the team to update the order.

I might be wrong because I haven't fully read Sarah's changes, but that's what I did on my end.

curtwl (dismissed):
lgtm, reordering worked as expected

curtwl (approved):
LGTM. I tried some edge cases like ties and everything looked good.

cherylli (approved):
works fine for me locally

PR #79, opened 2024-01-21
Add status to Voyage table and return in the /me endpoint

cherylli:
# Description

Add status to Voyage table and return in the /me endpoint

## Issue link

Fixes # (issue)

## Type of change

- [x] Non-breaking change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

# Checklist:


prisma/schema.prisma
curtwl, line 1: Should Voyage use the same VoyageStatus model as VoyageTeam? I was initially a bit confused since some of the statuses only apply to voyages, others only to voyage teams/members.

jenny-alexander: @cherylli Is there any value in separating these statuses into two separate tables? One for voyage status and one for user status?
Seems 'inactive', 'dropped', and 'inactive' alludes to member status & 'upcoming' and 'locked' refers to voyage status.

cherylli: I think they can all use the same status table as they are all just statuses (just an id and name, maybe description), it would be more confusing if we have a separate status table for everything

There's actually 3 things with statuses now and I think there would be more, another thing we could do is add a column type to the status table

Do you think it would be easier to just have a status table for everything separately? If we already have 100 tables, I guess it's not so much difference compared to 103 tables. 

Another option is to use ENUM but it's less flexible

jenny-alexander (approved):
Looks good and approved.

curtwl (approved):
lgtm

jenny-alexander:
[Update /me endpoint](https://trello.com/c/jRm0QZ5n/105-update-me-endpoint)

PR #88, opened 2024-01-26
Chore/delete tokens

curtwl:
# Description

Runs a cron job to delete EmailVerificationTokens and ResetTokens every two weeks.

For ease of testing, you can put this line in tasks.service.ts in place of the existing @Cron line to run the cron every 10 seconds.
```
@Cron('*/10 * * * * *')
```

Generate a test set of tokens with the following SQL. The EmailVerificationTokens are 3 weeks old and should be deleted. The ResetTokens are 1 week old and should not be deleted.
```
insert into
  "EmailVerificationToken" ("userId", token, "createdAt", "updatedAt")
select
  id,
  'verification_token_' || id,
  current_timestamp - interval '3 weeks',
  current_timestamp
from
  "User"
where
  not exists (
    select
      1
    from
      "EmailVerificationToken"
    where
      "EmailVerificationToken"."userId" = "User".id
  );

insert into
  "ResetToken" ("userId", token, "createdAt")
select
  id,
  'reset_token_' || id,
  current_timestamp - interval '1 week'
from
  "User"
where
  not exists (
    select
      1
    from
      "ResetToken"
    where
      "ResetToken"."userId" = "User".id
  );
```

## Issue link

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Monitored DB to verify that tokens older than two weeks are deleted and tokens newer than two weeks are not deleted.

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "@nestjs/jwt": "^10.1.1",
     "@nestjs/passport": "^10.0.2",
     "@nestjs/platform-express": "^10.0.0",
+    "@nestjs/schedule": "^4.0.0",
     "@nestjs/swagger": "^7.1.11",
     "@prisma/client": "^5.3.1",
     "bcrypt": "^5.1.1",
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -14,6 +14,8 @@ import { TechsModule } from "./techs/techs.module";
 import { FeaturesModule } from "./features/features.module";
 import { IdeationsModule } from "./ideations/ideations.module";
 import { JwtAuthGuard } from "./auth/guards/jwt-auth.guard";
+import { ScheduleModule } from "@nestjs/schedule";
+import { TasksModule } from "./tasks/tasks.module";
 
 @Module({
     imports: [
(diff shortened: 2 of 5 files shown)


cherylli (commented):
Hi, it looks good at a glance but I think it's missing the updated package.json (new packages need to be installed)

cherylli (approved):
Works for me locally, but do you know why in `yarn.lock`, all the `https://registry.yarnpkg.com/` changed to `https://registry.npmjs.org/` ?

curtwl:
> Works for me locally, but do you know why in `yarn.lock`, all the `https://registry.yarnpkg.com/` changed to `https://registry.npmjs.org/` ?

I'm still not very used to yarn, npm is muscle memory for me...the problem with the missing updated package.json was because I thought I accidentally did an npm install, so I did a git restore package.json. Oops...

As for why all that changed in yarn.lock, I'm not sure. I went through my bash history and it turns out that I did *not* do an npm install...

cherylli:
> in yarn.lock, I'm not sure. I went through my bash history and

I see. I don't think there will be an issue, since https://registry.yarnpkg.com/ redirects to https://registry.npmjs.org/ anyway, but __*if*__ there's any deployment issue we might have to revert this or fix it

jenny-alexander (approved):
LGTM! Added tokens via SQL and only reset tokens remained after running cron job.

PR #97, opened 2024-02-02
Authorization: Custom role/permission guards

cherylli:
# Description

Added roles and permissions custom guards and decorators. More roles and permissions can be added in future

As discussed on discord,

1. I don't think @Roles(Voyager) is really needed as OWN_TEAM permission guard works with restricting users to only access their own voyage regardless of their roles (which also works with previous voyages, even if they don't have the voyager role anymore, since voyager role = active voyage participant). But I think there's still value in keeping the voyager role as we can pull everyone with "voyager" role from the database easily

2. User role is probably redundant as as every registered user would be a "USER"

Manual seeding will be required after approval and successful database migration

We will need to update other routes/controllers when this is approved. Also remove `validateLoggedInAndTeamMember` in `global.service`. 

I have updated `users/teams` as an example of how this works

## Type of change

# How Has This Been Tested?

Tested locally on swagger

# Checklist:


prisma/seed/users.ts
curtwl, line 1: I got "Property 'userRole' does not exist on type 'PrismaClient<PrismaClientOptions, never, DefaultArgs>'." here...looks like it was renamed to VoyageRole?

cherylli: yes, not sure why it's not updated in the code! 
This is what I have locally `import VoyageRoles from "./data/voyage-roles";`

cherylli: actually, there are 2 roles now, roles and voyage-roles, did you run `prisma db push`?
roles is the newly added table for "chingu roles", the existing roles are "voyage roles"

cherylli: It looks like prisma.schema is not committed, will check later

cherylli:
It should be working again @curtwl, thanks for catching that. Please let me know if there's any other issues

jenny-alexander (approved):
LGTM

curtwl (approved):
Looks good!

PR #109, opened 2024-02-27
Feature/resources e2e

curtwl:
# Description

Add e2e tests for resources controllers. 

## Issue link

Fixes # (issue)

## Type of change

- [x] Tests

# How Has This Been Tested?

`yarn test:e2e resources`

# Checklist:


test/resources.e2e-spec.ts
```diff
@@ -0,0 +1,320 @@
...
+            },
+        },
+        select: {
+            userId: true,
+            voyageTeamId: true,
+        },
+    });
+};
+
+describe("ResourcesController (e2e)", () => {
```
cherylli, line 44: could you please categories the tests by endpoints, you can look at auth, techs e2e tests for example

cherylli (changes requested):
Could you structure the tests like 
describe (POST /voyages/:teamId/resources)
it(should return 201 if successful)
describe (should return 400 if request body is incomplete)
 it(missing title)
 it(missing url)
it (should return 401 if user is unauthorized)
etc

try to go through all responses we have on swagger

test/resources.e2e-spec.ts
```diff
@@ -0,0 +1,424 @@
...
+        .expect(200);
+
+    return extractResCookieValueByKey(
+        res.headers["set-cookie"],
+        "access_token",
+    );
+};
+
+const findVoyageTeamId = async (email: string, prisma: PrismaService) => {
+    return await prisma.voyageTeamMember.findFirst({
```
cherylli, line 31: await is redudant here, and line 45 (sorry my IDE is giving me a warning and I don't like seeing it 😂)

cherylli (changes requested):
Looks good, but could you please add return code for all the `it` blocks? 
maybe like it `should return 201` ... or `should create a new resource and return 201`

I also suggest adding more `expect`s to check database records before and after operation
- for create, check that the new entry is in the database 
- for delete, check the number of records after deletion matches number of records before - number of records deleted
- for update, check that it's actually updated

Thanks!

cherylli:
The tests are good, but it feels they are not consistent with other tests since we have all the 401 also included in specific endpoints in the other tests

JoshuaHinman (approved):
Tests passed.  Looks good!

cherylli:
> > The tests are good, but it feels they are not consistent with other tests since we have all the 401 also included in specific endpoints in the other tests. I'm not sure if it's enough to request a change, what do you think?
> 
> I moved things around to make it simpler to have two beforeEachs with the different user logins, but now that I think about it, I could probably just call a helper function that logs out the first user and logs in the second for the 401 tests...

Thanks. You could look at one of these or add what you need https://github.com/chingu-x/chingu-dashboard-be/blob/dev/test/utils.ts

cherylli (approved):
this looks good to me, thanks!

timDeHof (approved):
Good job! All unit tests and e2e tests passed.

PR #112, opened 2024-03-07
Add role and permission guard decorators to some existing routes

cherylli:
# Description

Add role and permission guard decorators to all routes except sprints, resources, and part of ideations, as I would like to look at adding additional permissions for these routes.

TODO - for future, not in this PR
https://github.com/chingu-x/chingu-dashboard-be/pull/112#issuecomment-1988547553

## Issue link

Fixes # (issue)

## Type of change

- [ ] Tests

# How Has This Been Tested?

image

# Checklist:


siasktv (approved):
Wow, this look like a massive amount of work. All tests passed!

src/forms/forms.controller.ts
```diff
@@ -28,6 +31,7 @@ export class FormsController {
         type: FormResponse,
         isArray: true,
     })
+    @Roles(AppRoles.Admin)
```
timDeHof, line 34: does role guard need to to be added under Roles decorator?
```suggestion
    @Roles(AppRoles.Admin)
    @UseGuard(RoleGuard)
```

src/teams/teams.controller.ts
```diff
@@ -31,7 +31,7 @@ export class TeamsController {
     constructor(private readonly teamsService: TeamsService) {}
 
     @ApiOperation({
-        summary: "Gets all voyage teams.",
+        summary: "[Roles: Admin] Gets all voyage teams.",
```
timDeHof, line 34: Do we have checklist for what is required when writing for swagger?

timDeHof (changes requested):
do we have tests for the guards?

src/teams/teams.controller.ts
cherylli: Not atm the moment, but I can write one if you think it would be helpful

cherylli:
> do we have tests for the guards?

yes we should have e2e tests on each endpoint for the permission, but I think we might need to go through each of them and add more soon, as i'm thinking to add more permissions

But I don't think we have tests for the role guard

cherylli:
I decided to only add role check for forms (also did some refactoring while i'm there) because
- ideations: will need to redo to match our current format/standard, too much for this PR
- teams: has no tests yet so it would be another PR itself to add the tests

Resources, sprints, users endpoints I have not touched so I don't intervere with the tests Josh, Curt, and Piero are working on - will need to add decorators to them - also because I want to add more permissions for these endpoints

image

cherylli:
back to draft again - I want to add missing swagger docs to the forms endpoint

cherylli:
ready for review again
image
image

cherylli:
new webhook test

test/utils.ts
timDeHof, line 1: Great idea of moving the utility functions to a separate file!

timDeHof (commented):
I like how you moved the utility functions to a separate file. We should properly do something similar to the function in the other test files.

curtwl (approved):
Successfully gave 403 for non-admin logged in user

timDeHof (approved):
great job! looks really clean.

PR #115, opened 2024-03-08
Features/forms response

siasktv:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Issue link

Fixes # (issue)

## Type of change

- [ ] Tests

# How Has This Been Tested?

yarn test:e2e forms

# Checklist:


prisma/migrations/20240305172254_1222/migration.sql
cherylli, line 1: why is there a migration? Did you change the database schema?

cherylli (changes requested):
hi, I get this response for sprints

image

which is not the desired response, refering to trello, we want something like this

image

src/forms/forms.service.ts
```diff
@@ -37,16 +41,23 @@ export const formSelect = {
...
 
-    getAllForms() {
-        return this.prisma.form.findMany({
+    async getAllForms() {
+        const data = await this.prisma.form.findMany({
             select: formSelect,
         });
+
+        const dataToReturn = data.map(({ questions, ...i }) => ({
+            ...i,
```
timDeHof, line 56: Maybe add a descriptive name to `i` in `...i`. It's hard read what it is.

timDeHof (changes requested):
I think the `i` at 'line 56 in form.service.ts' should be change to something more descriptive to make it easier to read.

prisma/migrations/20240305172254_1222/migration.sql
siasktv: Hi Cheryl, I have a few questions. I just modified the /forms route I'm not sure why are you getting that response for sprints? Will check as well. And it is questions == forms? because in the forms route the response was questions I modified it to be subquestions instead. Is that correct?

src/forms/forms.service.ts
siasktv: Hi Tim, thank you for the feedback! That's true will modify it as well.

cherylli:
not sure if this is ready for review again since I see a new push but no comments with it.

There are still a few issues
- subquestions are also appeared under questions, e.g. question id 18 is appeared under both. could you please remove them from questions
- is it possible to hide `subquestions` when there's no subquestions? `"subQuestions": []`
- will also need to update the sample response on swagger

cherylli:
marking this as draft as changes are required

siasktv:
> not sure if this is ready for review again since I see a new push but no comments with it.
> 
> There are still a few issues
> 
> * subquestions are also appeared under questions, e.g. question id 18 is appeared under both. could you please remove them from questions
> * is it possible to hide `subquestions` when there's no subquestions? `"subQuestions": []`
> * will also need to update the sample response on swagger

I think now is ready for review :)

src/forms/forms.response.ts
```diff
@@ -66,6 +66,38 @@ class Response {
     responseGroupId: number;
 }
 
+class subQuestion {
```
cherylli, line 69: class naming convention is PascalCase

cherylli: we also don't want to repeat a lot of code. This block of code is exactly the same as Question, without subQuestion

Easiest way is to use `Question extends SubQuestion` 


```
class SubQuestion {
    @ApiProperty({ example: 1 })
    id: number;

    @ApiProperty({ example: 1 })
    order: number;

    @ApiProperty()
    inputType: InputType;

    @ApiProperty({ example: "Changes to be made for the next sprint?" })
    text: string;

    @Optional()
    @ApiProperty({
        example:
            "Share your thoughts on what could be changed for the next sprint",
    })
    description: string;

    @ApiProperty({ example: false })
    answerRequired: boolean;

    @Optional()
    @ApiProperty({ example: null })
    multipleAllowed: boolean;

    @Optional()
    @ApiProperty({ example: null })
    optionGroup: OptionGroup;
}

class Question extends SubQuestion {
    @Optional()
    @ApiProperty({ isArray: true })
    subQuestions: SubQuestion;
}

```

siasktv: thank you for your insight. After doing some research around, I understand that a "sub-question" is considered a part of a "question". Therefore, it does seem more appropriate to have Question as the base class and SubQuestion as a subclass? Whoever when I tried this approach I get an error on can't access SubQuestion before initializing. I was thinking if it is better to make a base class for both? 
```
class Question {
    @ApiProperty({ example: 1 })
    id: number;

    @ApiProperty({ example: 1 })
    order: number;

    @ApiProperty()
    inputType: InputType;

    @ApiProperty({ example: "Changes to be made for the next sprint?" })
    text: string;

    @Optional()
    @ApiProperty({
        example:
            "Share your thoughts on what could be changed for the next sprint",
    })
    description: string;

    @ApiProperty({ example: false })
    answerRequired: boolean;

    @Optional()
    @ApiProperty({ example: null })
    multipleAllowed: boolean;

    @Optional()
    @ApiProperty({ example: null })
    optionGroup: OptionGroup;

    @Optional()
    @ApiProperty({ isArray: true })
    subQuestions?: SubQuestion[];
}

class SubQuestion extends Question {
    @Optional()
    @ApiProperty({ example: null })
    optionGroupId: number | null;

    @ApiProperty()
    parentQuestionId: number;

    @ApiProperty()
    createdAt: Date;

    @ApiProperty()
    updatedAt: Date;
}
```

cherylli: you can if you want to, but basically question should be the same as subquestion but with `subQuestions` field

so maybe you can do something like 

```
BaseQuestion {
  ...
}

Question extends BaseQuestions {
  subQuestions
}

SubQuestion extends BaseQuestions {}

```

if you want to include createdAT/updatedAt for subqestion we should have it for questions too
And you won't need "parentQuestionId" in subquestion as it would be a child node of a question. I'm not sure if we will ever need to pull a subquestion by itself

I think we will eventually need createdAt and updatedAt for both because it would be handy to show question last update in the admin dashboard

src/forms/forms.response.ts
```diff
@@ -97,8 +97,23 @@ class Question {
...
-    responses: Response;
+    @ApiProperty()
+    createdAt: Date;
+
+    @ApiProperty()
+    updatedAt: Date;
+}
+
+class SubQuestion extends BaseQuestion {
+    // No need for parentQuestionId in SubQuestion
```
cherylli, line 108:
```suggestion
   
```

src/forms/forms.response.ts
```diff
@@ -97,8 +97,23 @@ class Question {
...
+    // @ApiProperty()
+    // parentQuestionId: number;
```
cherylli, line 110:
```suggestion
    
```

src/forms/forms.response.ts
```diff
@@ -97,8 +97,23 @@ class Question {
...
+}
+
+class Question extends BaseQuestion {
+    @Optional()
+    @ApiProperty({ isArray: true, type: SubQuestion })
```
cherylli, line 115: should this be type:Question?

src/forms/forms.service.ts
```diff
@@ -35,31 +39,102 @@ export const formSelect = {
...
+
+            form.questions = filteredQuestions;
+        }
+
+        return data;
     }
 
     async getFormById(formId: number) {
-        const form = await this.prisma.form.findUnique({
+        const data = await this.prisma.form.findUnique({
```
cherylli, line 111: i think this should be `form` as the returned object is a form object
you can see an example in the prisma docs

cherylli (commented):
thanks, just looked at the code quickly before bed, will test it in action tomorrow 😊

src/forms/forms.response.ts
cherylli, line 110:
```suggestion
```

cherylli:
thanks, tested on swagger, everything works and all existing tests passed

src/forms/forms.response.ts
siasktv: sure, no problem! will finish with the changes and update PR later today

src/forms/forms.service.ts
```diff
@@ -35,18 +39,72 @@ export const formSelect = {
...
+} as Prisma.FormSelect;
 
 @Injectable()
 export class FormsService {
     constructor(private prisma: PrismaService) {}
 
-    getAllForms() {
-        return this.prisma.form.findMany({
+    async getAllForms() {
+        const data = await this.prisma.form.findMany({
```
cherylli, line 81: should we use something more descriptive than `data`?

siasktv: I can switch it to: forms?

cherylli:
> * will also need to update the sample response on swagger

it seems the swagger sample response doesn't match the actual response
image

siasktv:
Hi @cherylli I checked swagger sample response and it seems it matches with the response. I couldn't find the error, could you please be more specific? 
image

cherylli:
thanks. it looks fine now. I must be looking at an older version sorry

cherylli (approved):

curtwl (approved):
Looks good!

timDeHof (approved):
great job! All tests passed

PR #116, opened 2024-03-08
Feat/delete users tokens

timDeHof:
# Description

This PR adds a DELETE endpoint to auth services that will revoke an user's refresh token when given a user's id or email.This endpoint has role guard attached so that only users with admin role can initiate the endpoint.  

Also this PR tests the new functionality of revoking a refresh token. The tests cover scenarios where the refresh token is successfully revoked, where the email is invalid, and where the user is not permitted to revoke the token. These tests ensure that the revoking of refresh tokens is working correctly and that the appropriate responses are returned based on the different scenarios.

In addition, swagger UI api has been updated to allow for users to test the new endpoint.

## Issue link

Fixes # (issue) "Create endpoint to revoke refresh token" from trello

## Type of change

- [ ] Tests

# How Has This Been Tested?

You can run `yarn test:e2e` or `yarn test:e2e:docker`, then look for `auth.e2e-spec.ts` to pass.
Screenshot 2024-03-08 at 1 05 13 PM
Screenshot 2024-03-08 at 1 04 41 PM

# Checklist:


src/auth/auth.controller.ts
```diff
@@ -196,6 +202,38 @@ export class AuthController {
...
+        type: BadRequestErrorResponse,
+    })
+    @ApiResponse({
+        status: HttpStatus.FORBIDDEN,
+        description: "user doesn't have permission to preform operation",
+        type: ForbiddenErrorResponse,
+    })
+    @HttpCode(HttpStatus.OK)
+    @Roles(AppRoles.Admin)
+    @UseGuards(RolesGuard)
```
cherylli, line 231: @useGuards(RolesGuard) shouldn't be needed, as we have a app level rolesGuard 
(`src/app.module.ts`)

src/auth/auth.controller.ts
```diff
@@ -34,6 +36,10 @@ import { JwtAuthGuard } from "./guards/jwt-auth.guard";
 import { JwtRefreshAuthGuard } from "./guards/jwt-rt-auth.guard";
 import { Public } from "../global/decorators/public.decorator";
 import { AT_MAX_AGE, RT_MAX_AGE } from "../global/constants";
+import { RevokeRTDTo } from "./dto/revoke-refresh-token.dto";
```
cherylli, line 39:
```suggestion
import { RevokeRTDto } from "./dto/revoke-refresh-token.dto";
```

test/auth.e2e-spec.ts
```diff
@@ -421,6 +421,86 @@ describe("AuthController e2e Tests", () => {
...
+                data: {
+                    refreshToken: null,
+                },
+            });
+
+            await request(app.getHttpServer())
+                .delete(revokeRTUrl)
+                .set("Cookie", [access_token, refresh_token])
+                .send({ email: "l.castro@outlook.com" })
+                .expect(200);
```
cherylli, line 447: should we add an expect (refreshtoken tobe null) in the database after revoke?

test/auth.e2e-spec.ts
```diff
@@ -421,6 +421,86 @@ describe("AuthController e2e Tests", () => {
...
+                .delete(revokeRTUrl)
+                .set("Cookie", [access_token, refresh_token])
+                .send({
+                    userId: "c4daf07c-8dde-4a43-bfa4-a6fd49762dd5",
+                    email: "l.castro@outlook.com",
+                })
+                .expect(400);
+        });
+
+        it("should return 403 if user is not permitted", async () => {
```
cherylli, line 485: I like that you included this test, we should definitely go back to the older tests to include these once we added other permissions

cherylli (changes requested):
this PR is very well done. Thank you

test/auth.e2e-spec.ts
timDeHof: That would be a good idea. But when I tried to add `.expect(refreshToken).toBeNull()`, I got an error saying "Property 'toBeNull' does not exist on type 'Test'"

timDeHof: I added a separate test that checks that the user's refresh token in null

cherylli:
image
I think this is not accurate, since we can either use email or userID, maybe it should be `auth/refresh/revoke` or `auth/revoke-refresh-token`

You can look at this for reference https://restfulapi.net/resource-naming/

------------------
image
I got this when user is not found in the database, it's not very descriptive

------------------
200 - delete token successful works, but there's no response message
image

cherylli (changes requested):
image
I think this is not accurate, since we can either use email or userID, maybe it should be `auth/refresh/revoke` or `auth/revoke-refresh-token`

You can look at this for reference https://restfulapi.net/resource-naming/

------------------
image
I got this when user is not found in the database, it's not very descriptive

------------------
200 - delete token successful works, but there's no response message
image

src/auth/auth.service.ts
```diff
@@ -133,6 +135,54 @@ export class AuthService {
...
+                        id: userId,
+                    },
+                    data: {
+                        refreshToken: null,
+                    },
+                });
+            } catch (error) {
+                throw new NotFoundException({
+                    statusCode: 404,
+                    message: `User by "${userId}" cannot be found.`,
```
cherylli, line 159:
```suggestion
                    message: `User by '${userId}' cannot be found.`,
```

just look better on swagger 
image
vs
image

src/auth/auth.service.ts
```diff
@@ -133,6 +135,54 @@ export class AuthService {
...
+                        email: email,
+                    },
+                    data: {
+                        refreshToken: null,
+                    },
+                });
+            } catch (error) {
+                throw new NotFoundException({
+                    statusCode: 404,
+                    message: `User by "${email}" cannot be found.`,
```
cherylli, line 175:
```suggestion
                    message: `User by '${email}' cannot be found.`,
```

cherylli (approved):
thanks for the changes!

cherylli (approved):

JoshuaHinman (approved):
e2e tests passed , swagger responses work, code looks great!

PR #156, opened 2024-05-22
Fix form responses giving error and not inserting values when the boolean value is false

cherylli:
# Description

When response is false, the validation pipe returns an error and the boolean value is also not inserted into the database properly

This is a new PR after renaming #154 as the branch name was causing issue for some members

## Issue link

[Fixes # (issue)](https://app.clickup.com/t/86b0e28ea)

## Type of change

- [ ] Tests

# How Has This Been Tested?

ran all the tests, send request from swagger and verified the values in the database 

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ Another example [here](https://co-pilot.dev/changelog)
 - Fix a bug in PATCH /meetings/{meetingId}/forms/{formId} where it's not accepting an array of responese (updated validation pipe, service, and tests) ([#121](https://github.com/chingu-x/chingu-dashboard-be/pull/121))
 - Fix unit tests where mocked req doesn't match new CustomRequest type ([#122](https://github.com/chingu-x/chingu-dashboard-be/pull/122))
 - Fix bug with reading roles after reseeding causes the db to not recognize the tokens stored by the user's browser ([#134](https://github.com/chingu-x/chingu-dashboard-be/pull/134))
+- Fix form responses giving error and not inserting values when the boolean value is false ([#154](https://github.com/chingu-x/chingu-dashboard-be/pull/154))
 
 ### Removed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ Another example [here](https://co-pilot.dev/changelog)
 - Fix a bug in PATCH /meetings/{meetingId}/forms/{formId} where it's not accepting an array of responese (updated validation pipe, service, and tests) ([#121](https://github.com/chingu-x/chingu-dashboard-be/pull/121))
 - Fix unit tests where mocked req doesn't match new CustomRequest type ([#122](https://github.com/chingu-x/chingu-dashboard-be/pull/122))
 - Fix bug with reading roles after reseeding causes the db to not recognize the tokens stored by the user's browser ([#134](https://github.com/chingu-x/chingu-dashboard-be/pull/134))
+- Fix form responses giving error and not inserting values when the boolean value is false ([#154](https://github.com/chingu-x/chingu-dashboard-be/pull/154))
```
cherylli, line 70:
```suggestion
- Fix form responses giving error and not inserting values when the boolean value is false ([#156](https://github.com/chingu-x/chingu-dashboard-be/pull/156))
```

Ajen07 (approved):
All tests works fine . No error when the response is false and database is correctly updated for false values

JoshuaHinman (approved):
Tests passed and boolean false verified with swagger and prisma studio.  One thing I was curious about: in the `responseDtoToArray` function, when pushing to the responsesArray, what do the spread operators do?

cherylli:
> Tests passed and boolean false verified with swagger and prisma studio. One thing I was curious about: in the `responseDtoToArray` function, when pushing to the responsesArray, what do the spread operators do?
@JoshuaHinman 
Thanks for asking the question, I don't remember why I did that it seemed natural to me at the point, but then I did some research, it seems, with the spread operator, you can conditionally set the values
without spread operator, all have to be included in the push vs the alternative

```
for (const index in responses) {
    if (responseIndex.includes(index)) {
        responses[index].forEach((v: FormResponseDto) => {
            responsesArray.push({
                questionId: v.questionId,
                text: v.text,
                numeric: v.numeric,
                boolean: v.boolean,
                optionChoiceId: v.optionChoiceId,
            });
        });
    }
}
```

which probably did not matter if we have all these values inserted into the database (including nulls), I was trying to convert into the same structure as the input, but I think empty values are undefined coming from the res body, so I don't think we could just leave it as `undefined` as we can insert null into the database but not undefined(not tested, just my theory)

JoshuaHinman:
>  it seems, with the spread operator, you can conditionally set the values
Thank you for explaining this - I don't recall seeing it before

PR #163, opened 2024-06-12
Unit test/voyages

cherylli:
# Description

Add unit test for voyages controller and voyages service. 
While adding the prismaMock, strict null check was required to be set to true to avoid circular imports
https://www.prisma.io/docs/orm/prisma-client/testing/unit-testing#mocking-the-prisma-client

As a result, a lot of the files has to be updated.

## Issue link

[Fixes # (issue)](https://app.clickup.com/t/86b0ucwrr)

## Type of change

- [x] Tests

# How Has This Been Tested?

image

# Checklist:


src/sprints/sprints.service.ts
```diff
@@ -71,9 +71,12 @@ export class SprintsService {
                 },
             },
         });
-        return sprintsByTeamId?.voyage?.sprints?.filter(
+
+        const sprintId = sprintsByTeamId?.voyage?.sprints?.filter(
             (s) => s.number === sprintNumber,
         )[0]?.id;
+
+        return sprintId;
```
Ajen07, line 79: We can return only ` null ` when  `sprintId` is not found rather than having both types `null` and `undefined`

```
   const sprintId = sprintsByTeamId?.voyage?.sprints?.filter(
            (s) => s.number === sprintNumber,
        )[0]?.id || null;
```

Ajen07 (changes requested):

src/sprints/sprints.service.ts
cherylli: I'd rather keep it undefined if it's undefined, null and undefined are different.

Although it looks like we can remove null

Ajen07: Yeah  we can remove either of one .

curtwl (approved):
lgtm!

Ajen07 (approved):
Great job ! All the changes looks good.

PR #166, opened 2024-06-16
Feature/checkin form endpoints

curtwl:
# Description

Retrieves check-in form responses given a sprintNumber, voyageId, userId, or teamId query param or any combination of them.
I also added a global function that does the basic if !itemFound throw 404 logic given a table name and key/value. 

## Issue link

https://app.clickup.com/t/86azuu2yc

Fixes # (issue)

## Type of change

- [ ] Tests

# How Has This Been Tested?

New tests and manual verification with Swagger that errors are handled as expected.

# Checklist:


src/sprints/sprints.response.ts
```diff
@@ -222,3 +222,32 @@ export class CheckinSubmissionResponse {
...
+
+    @ApiProperty({ example: "2023-11-30T06:47:11.694Z" })
+    responseGroupId: number;
+
+    @ApiProperty({ example: "2023-11-30T06:47:11.694Z" })
+    createdAt: Date;
+
+    @ApiProperty({ example: "2023-11-30T06:47:11.694Z" })
+    updatedAt: Date;
+}
```
Ajen07, line 253: adminComments , feedbackSent , responseGroupId  examples can be updated to their respective types rather having a date-time format for each one

Ajen07 (changes requested):

JoshuaHinman (commented):
When I ran unit tests, these lines in `sprints.contoller` threw errors:
image
I changed the paths form `src/` to `../` to get them working

curtwl:
> When I ran unit tests, these lines in `sprints.contoller` threw errors: image I changed the paths form `src/` to `../` to get them working

Strange, both ways work for me... I've updated them to use ../ instead of src/

Ajen07 (approved):
Great Job ! All the changes looks good.

cherylli (approved):

cherylli (changes requested):
accidentally pressed the approved button

curtwl:
> 1. Is it possible also populate the responses? In the response I can only see `responseGroupId:1`,  this doesn't really allow the admin to see the responses when they pull the submission.
> 
> ===========
> 
> 2. I also see something weird, I get the same result regardless of what sprintNumber I put, I always get sprintId: 1 (there's only 1 check in form in my database for sprint 1)
> 
> image
> 
> =============
> 
> 3. when `voyageNumber` is not given, sprint Number should not work, as we don't know sprint x of what voyage, I'm getting records for sprintId 2. Please note that sprint number and sprint id are different.

I see what you mean, will get that fixed up.

JoshuaHinman (approved):
All tests passed, and I was able to query check-ins using each param - nice work!
I have a quick question about one line from `sprints.service.ts` that I was confused about:
image
 `queryObjectStringified` is an array of strings, right?  I was wondering why/how it extracts key and value from a string

curtwl:
> All tests passed, and I was able to query check-ins using each param - nice work! I have a quick question about one line from `sprints.service.ts` that I was confused about: image `queryObjectStringified` is an array of strings, right? I was wondering why/how it extracts key and value from a string

I realized my description in the comment is actually not quite right. JS internally stores the key names as strings already, and this function just puts them (and their values) into an array as they are just to make it easier to do the switch statement. I couldn't think of any way other than that or a really long ternary chain, but if there's a cleaner way to do it, I'd love to know!

cherylli (changes requested):
I hope this won't hard to fix but this combination returns the wrong thing
image


It returns checkin form submission that doesn't belong to `teamId=15`, it seems to ignore teamId,

I created another check in form teamId=8 and both were returned regardless of what teamId I entered

and this for example - 
image

I think it should not return anything, or an error saying user is not in teamId 8, since the teamId is not 8. If there's any contraction e.g. if team 8 is not in v46 we should also return that error

===================================
I also think if the query params are correct but there's no check in form found we should use status 200 with an empty array

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404

> The HTTP 404 Not Found response status code indicates that the server cannot find the requested resource. Links that lead to a 404 page are often called broken or dead links and can be subject to [link rot](https://en.wikipedia.org/wiki/Link_rot).

It would be for cases like teamId does not exist then url would be invalid or broken

curtwl:
> I hope this won't hard to fix but this combination returns the wrong thing image
> 
> It returns checkin form submission that doesn't belong to `teamId=15`, it seems to ignore teamId,
> 
> I created another check in form teamId=8 and both were returned regardless of what teamId I entered
> 
> and this for example - image
> 
> I think it should not return anything, or an error saying user is not in teamId 8, since the teamId is not 8. If there's any contraction e.g. if team 8 is not in v46 we should also return that error
> 
> =================================== I also think if the query params are correct but there's no check in form found we should use status 200 with an empty array
> 
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
> 
> > The HTTP 404 Not Found response status code indicates that the server cannot find the requested resource. Links that lead to a 404 page are often called broken or dead links and can be subject to [link rot](https://en.wikipedia.org/wiki/Link_rot).
> 
> It would be for cases like teamId does not exist then url would be invalid or broken
> 
> For example, voyage 46, sprint 6, this resource definitely exists but just no checkin form submissions

I thought from the clickup description (with phrasing like "get all for a particular voyage (all sprints)") that the route was meant to take one of four options (including sprintNumber+voyageNumber as one separately), not combinations of them. So that was by design -- only the first key is used and others are ignored, rather than making the user try again.

Changing 404 to 200 for no check-in forms found but valid params should be easy to fix with some of the global functions. Wasn't sure how to handle that. So it should just return an empty array and frontend will take it from there?

cherylli:
> I thought from the clickup description (with phrasing like "get all for a particular voyage (all sprints)") that the route was meant to take one of four options (including sprintNumber+voyageNumber as one separately), not combinations of them. So that was by design -- only the first key is used and others are ignored, rather than making the user try again.

user + voyage number + sprint number should be a valid choice. voyage number + sprint number + team id should also be a valid choice. Sorry I forgot to list this cases. Either way, I don't think it's a good idea to just "ignore" user inputs, we should return an error message saying the combination is not allowed, instead of returning some unexpected results. 

I'm not sure what's the best way to handle it, as a user it feels like i'd be able to search/filter by a combination when the options are presented this way. I guess we sort of deviated a bit when we decide to use the query params.

I think what you're doing now is great as it can handle both cases (get all... and get more specific), but just some common combinations don't seem to be working. 

> So it should just return an empty array and frontend will take it from there? 
yes they could handle cases for array length === 0

src/sprints/sprints.response.ts
```diff
@@ -222,3 +223,53 @@ export class CheckinSubmissionResponse {
...
+
+class SprintNumberWithVoyageNumber {
+    @ApiProperty({ example: 1 })
+    number: number;
+
+    @ApiProperty({ example: 46 })
+    voyage: VoyageNumber;
+}
+
+export class CheckinFormResponse {
```
cherylli, line 245: I think this is not used, so the response is still showing the old ones, 
image
I assume it's supposed to go here
image

cherylli (changes requested):
I like this approach, we could reuse the logic for other endpoints like fetching solo projects and other forms in future 👍
checked a few random cases and they were fine, all tests passed
just an unused response posted in comment

JoshuaHinman (changes requested):
I got correct responses with all params, including some combinations I tried.  However, there was one issue I stumbled on: when querying with sprint number 0, it returns all sprints.

src/sprints/sprints.response.ts
curtwl: Oh lol, I managed to confuse myself by making the name similar to CheckinSubmissionResponse

curtwl:
> I got correct responses with all params, including some combinations I tried. However, there was one issue I stumbled on: when querying with sprint number 0, it returns all sprints.

Fixed, now sprint numbers are restricted to 1-6

cherylli:
> > I got correct responses with all params, including some combinations I tried. However, there was one issue I stumbled on: when querying with sprint number 0, it returns all sprints.
> 
> Fixed, now sprint numbers are restricted to 1-6

that reminds me we need to fix create-checkin-form.dto.ts

cherylli (approved):
Thanks for the quick fix!

curtwl:
> > > I got correct responses with all params, including some combinations I tried. However, there was one issue I stumbled on: when querying with sprint number 0, it returns all sprints.
> > 
> > 
> > Fixed, now sprint numbers are restricted to 1-6
> 
> that reminds me we need to fix create-checkin-form.dto.ts
> 
> edit: should probably have asked before pushing my changes but I thought it might be easier if I change it quickly, and it's not one of the files you changed. 🤦‍♀️🙇‍♀️

I tried that locally a few days ago, but I noticed a complication, that someone on i.e. a v48 team can submit check-ins for any sprint id, and end up with check-ins in multiple voyages with that team.

cherylli:
> > > > I got correct responses with all params, including some combinations I tried. However, there was one issue I stumbled on: when querying with sprint number 0, it returns all sprints.
> > > 
> > > 
> > > Fixed, now sprint numbers are restricted to 1-6
> > 
> > 
> > that reminds me we need to fix create-checkin-form.dto.ts
> > edit: should probably have asked before pushing my changes but I thought it might be easier if I change it quickly, and it's not one of the files you changed. 🤦‍♀️🙇‍♀️
> 
> I tried that locally a few days ago, but I noticed a complication, that someone on i.e. a v48 team can submit check-ins for any sprint id, and end up with check-ins in multiple voyages with that team.

We probably need to add a check for that, currently we have check for the teamMemberId but not sprintId, we will probably need to check teamMemberId->voyageTeam->voyage->sprintId, but that's a lot more complicated than just restricting the DTO, so this can be its own PR (unless its actually simpler than I thought... 🤔)

JoshuaHinman (approved):
Retested - everything passed.  Excellent work!

PR #168, opened 2024-06-23
Test: e2e tests for features

Ajen07:
# Description

- e2e test for `features controller`

## Issue link

[Fixes # (issue)](https://app.clickup.com/t/86azr0q4j)

## Type of change

- [x] Tests

# How Has This Been Tested?

- tested with  `yarn test:docker`

# Checklist:

### DIFF
--- a/src/features/features.controller.ts
+++ b/src/features/features.controller.ts
@@ -26,6 +26,7 @@ import {
 import { Feature } from "./entities/feature.entity";
 import {
     BadRequestErrorResponse,
+    ForbiddenErrorResponse,
     NotFoundErrorResponse,
     UnauthorizedErrorResponse,
 } from "../global/responses/errors";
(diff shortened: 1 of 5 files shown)


src/features/features.controller.ts
```diff
@@ -196,7 +207,7 @@ export class FeaturesController {
         } else {
             throw new HttpException(
                 "user is unauthorized to perform this action",
```
cherylli, line 209:
```suggestion
                "Access denied: You do not have sufficient permissions to perform this action",
```

cherylli (dismissed):
everything looks fine, just one minor suggestion so the error response matches the forbidden error

Also just a note all forbidden stuff will need to be moved to our CASL authorization, but shouldn't affect the tests. In fact it's good to have these tests now, so we know we don't mess up when updating it

JoshuaHinman (approved):
All tests passed - looks good

src/features/features.controller.ts
Ajen07: Currently in some controllers there are no permissions for own_team , after the permissions are added then, we can add the forbidden tests to it.

PR #177, opened 2024-07-22
Feature: casl permission for resources

Ajen07:
# Description

- Add new permissions for Team Resources endpoint

## Issue link

[Fixes # (issue)](https://app.clickup.com/t/86azr0q42)

## Type of change

- [ ] Tests

# How Has This Been Tested?

- test with `yarn test:docker`

- tested with swagger

# Checklist:


src/resources/resources.service.ts
```diff
@@ -2,43 +2,39 @@ import {
...
         teamId: number,
     ) {
-        const { url, title } = createResourceDto;
-        const userId = req.user.userId;
-
         // make sure teamId exists
         await this.checkTeamExists(teamId);
 
-        const teamMember =
-            await this.globalService.validateLoggedInAndTeamMember(
```
cherylli, line 31: I like that you're replacing this as I plan to retire this in future as it requires us to query the database when we already have the information in `req.user`

src/resources/resources.service.ts
```diff
@@ -2,43 +2,39 @@ import {
...
-                teamId,
-                userId,
-            );
+        manageOwnVoyageTeamWithIdParam(req.user, teamId);
+        const { url, title } = createResourceDto;
+
+        const teamMemberId = req.user.voyageTeams.find(
```
cherylli, line 28: is there a possibility that teamMemberId is null/undefined? If yes we need to handle this, instead of passing a possible undefined below (line 50). 

Typescript says it can be `number | underfined` but `manageOwnVoyageTeamWithIdParam` should have ruled out that possiblity to be undefined? Unless the user is an admin which bypasses that check

Would it be beneficial to extract this function? Feels like it could be something useful for other services? It looks like there's another usage in features.service

cherylli (commented):
All tests passed. Everything else look good. I just have one question (commented in the code)

src/resources/resources.service.ts
Ajen07: Currently , I see that there is  a `getVoyageTeamMemberId` in the `global service` . So we can use that function to find the member Id , hence it will not be undefined anymore.

cherylli: @Ajen07 good find. I don't even remember this function😂

cherylli (approved):
👍🚀

JoshuaHinman (approved):
Confirmed authorized and not authorized responses in swagger, and all tests passed!

PR #188, opened 2024-08-22
End of Sprint 26 merge into main

cherylli:


prisma/production-seed/users.ts
```diff
@@ -0,0 +1,358 @@
...
+
+    //<editor-fold desc="admin account">
+    const adminUser = await prisma.user.create({
+        data: {
+            email: "admin@example.com",
+            password: await hashPassword("password"),
+            emailVerified: true,
+            firstName: "Admin",
+            lastName: "Surname",
+            avatar: "https://gravatar.com/avatar/3bfaef00e02a22f99e17c66e7a9fdd31?s=400&d=robohash&r=x",
```
timDeHof, line 16: I think if you want different generated robot, you have to change the hash for each user. Also I don't think it means anything but `r=x` is the gravatar images rating. We could probably just remove the rating, because the default rating is `g`, which allows for only g-rated images to be displayed on the website. 
[avatars request - gravatar docs](https://docs.gravatar.com/api/avatars/images/)

timDeHof (changes requested):
unit and e2e tests passed. I left a comment about the gravatar image that may want to make changes to. Also do we want to make the change to question 11 in the `checkin.ts`?

cherylli:
> unit and e2e tests passed. I left a comment about the gravatar image that may want to make changes to. Also do we want to make the change to question 11 in the `checkin.ts`?

I have added a ticket for both issues. We can just not do a push this sprint since we just pushed to main not long ago.

timDeHof (approved):
All unit & e2e tests pass on dev

JoshuaHinman (approved):
All tests passed

PR #193, opened 2024-08-31
Feat: CASL permission voyage sprints

Ajen07:
# Description

- Add  permissions for Voyage Sprint endpoint

## Issue link

[Fixes # (issue)](https://app.clickup.com/t/86azr0q42)

## Type of change

- [ ] Tests

# How Has This Been Tested?

- test with yarn `test:docker`
- tested with `swagger`

# Checklist:


src/sprints/sprints.controller.ts
```diff
@@ -125,19 +145,28 @@ export class SprintsController {
...
-    getMeetingById(@Param("meetingId", ParseIntPipe) meetingId: number) {
-        return this.sprintsService.getMeetingById(meetingId);
+    getMeetingById(
+        @Request() req: CustomRequest,
+        @Param("meetingId", ParseIntPipe) meetingId: number,
+    ) {
+        return this.sprintsService.getMeetingById(meetingId, req);
     }
 
+    @CheckAbilities({ action: Action.Create, subject: "Sprint" })
```
cherylli, line 169: I wonder if we should add new action `CreateSprintMeeting`, or `ManageSprintMeeting`
Action.Create, subject: "Sprint" sounds like create the actual sprints which should be only admin

src/sprints/sprints.controller.ts
```diff
@@ -247,26 +289,34 @@ export class SprintsController {
...
+        description: "forbidden - user does not have the required permission",
+        type: ForbiddenErrorResponse,
+    })
     @ApiParam({
         name: "meetingId",
         required: true,
         description: "voyage team meeting ID",
         example: 1,
     })
+    @CheckAbilities({ action: Action.Create, subject: "Sprint" })
```
cherylli, line 306: same as above

src/sprints/sprints.controller.ts
```diff
@@ -283,26 +333,34 @@ export class SprintsController {
...
+        description: "forbidden - user does not have the required permission",
+        type: ForbiddenErrorResponse,
+    })
     @ApiParam({
         name: "agendaId",
         required: true,
         description: "agenda ID",
         example: 1,
     })
+    @CheckAbilities({ action: Action.Update, subject: "Sprint" })
```
cherylli, line 350: same as above, maybe Action.ManageSprintMeetings - which would include all operations

src/sprints/sprints.controller.ts
```diff
@@ -319,20 +377,29 @@ export class SprintsController {
...
+    })
     @ApiParam({
         name: "agendaId",
         required: true,
         description: "agenda ID",
         example: 1,
     })
-    deleteMeetingAgenda(@Param("agendaId", ParseIntPipe) agendaId: number) {
-        return this.sprintsService.deleteMeetingAgenda(agendaId);
+    @CheckAbilities({ action: Action.Delete, subject: "Sprint" })
```
cherylli, line 394: same here

src/sprints/sprints.service.ts
```diff
@@ -273,91 +280,159 @@ export class SprintsService {
...
-        try {
-            const updatedMeeting = await this.prisma.agenda.update({
-                where: {
-                    id: agendaId,
-                },
-                data: {
-                    title,
-                    description,
-                    status,
+        const agenda = await this.prisma.agenda.findUnique({
```
cherylli, line 352: This is a good solution even though we added one extra database query. I'm not sure if there's another way to just have 1 query with the addition of authorisation

src/ability/ability.factory/ability.factory.ts
```diff
@@ -52,6 +52,7 @@ export class AbilityFactory {
                     in: user.voyageTeams.map((vt) => vt.memberId),
                 },
             });
+            can([Action.Manage], "Sprint");
```
cherylli, line 55: I think we should create another Action, as stated below - voyages shouldn't be able to manage sprint

cherylli (commented):
looks good so far, what do you think about adding a new Action for the forms?

src/sprints/sprints.controller.ts
Ajen07: I was thinking instead of creating a new action  we can create a new subject  for sprintMeeting

cherylli: That's a better idea if you can make it work but I wasn't sure, I thought the subject has to be tied to a prisma model.  

We do have a team meeting model

Ajen07: Yes we have a team meeting model and if we create a subject for it agenda and meeting related operations can be handled at one place.

timDeHof (approved):
All tests passed! everything works on swagger

timDeHof (dismissed):
All tests passed! everything works on swagger

src/ability/conditions/meetingOrAgenda.ability.ts
```diff
@@ -0,0 +1,68 @@
...
+        if (!agenda) {
+            throw new NotFoundException(
+                `Team Meeting Agenda (id:${agendaId}) not found`,
+            );
+        }
+
+        meetingOrAgendaTeamId = agenda.teamMeeting.voyageTeamId;
+    }
+
+    if (user.roles?.includes("admin")) return;
```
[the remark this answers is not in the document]
Ajen07: we have to go through the db queries because  in the above queries we are only checking if  a meeting or agenda exist or not for a given id as it is required for all roles.

src/sprints/sprints.service.ts
```diff
@@ -263,7 +271,23 @@ export class SprintsService {
             },
         });
     }
+    async helper(meetingId: number) {
```
cherylli, line 274: I think we should name this something more descriptive then `helper`

src/sprints/sprints.service.ts
```diff
@@ -273,91 +297,91 @@ export class SprintsService {
...
-            });
-            return updatedMeeting;
-        } catch (e) {
-            if (e.code === "P2025") {
-                throw new NotFoundException(`Invalid meetingId: ${meetingId}`);
-            }
-        }
+        await manageOwnTeamMeetingOrAgendaById({ user: req.user, meetingId });
+
+        const updatedMeeting = await this.prisma.teamMeeting.update({
```
cherylli, line 304: we can just return here (without making a new variable)

src/sprints/sprints.service.ts
```diff
@@ -273,91 +297,91 @@ export class SprintsService {
...
-        } catch (e) {
-            if (e.code === "P2003") {
-                throw new BadRequestException(
-                    `Invalid meetingId: ${meetingId}`,
-                );
-            }
-        }
+        await manageOwnTeamMeetingOrAgendaById({ user: req.user, meetingId });
+
+        const newAgenda = await this.prisma.agenda.create({
```
cherylli, line 326: we can also remove the variable creation here

src/sprints/sprints.service.ts
```diff
@@ -273,91 +297,91 @@ export class SprintsService {
...
+            where: {
+                id: agendaId,
+            },
+            data: {
+                title,
+                description,
+                status,
+            },
+        });
+        return updatedMeeting;
```
cherylli, line 353: redundant variable creation

test/sprints.e2e-spec.ts
```diff
@@ -228,27 +287,62 @@ describe("Sprints Controller (e2e)", () => {
...
+            const { access_token, refresh_token } = await loginAndGetTokens(
+                "jessica.williamson@gmail.com",
+                "password",
+                app,
+            );
             const meetingId = 9999;
             return request(app.getHttpServer())
                 .get(`/voyages/sprints/meetings/${meetingId}`)
-                .set("Cookie", accessToken)
+                .set("Cookie", [access_token, refresh_token])
```
cherylli, line 298: no change required here but just curious, is refresh_token needed? I wonder why it's added now

cherylli (commented):
Everything looks good just a few very minor changes/comments
I like how you have included cases for both non-voyages and other team in the tests for forbidden access, and also updated all the swagger descriptions

cherylli:
I have also ran all tests and they all passed locally

src/sprints/sprints.service.ts
Ajen07: Actually we don't need this function , I created this to test something but forgot to remove it 😅

cherylli (dismissed):
nice work!

src/ability/ability.factory/ability.factory.ts
timDeHof, line 1: I see that checking `voyageTeams` and mapping their IDs is repeated multiple times. Also  mapping over `user.voyageTeams` multiple times might not be efficient if the list is large. I think it can be optimized by refactoring the voyageTeam's Ids and voyageTeamMember's Ids by creating helper functions to reduce duplication and store the mapped values in variables.

timDeHof, line 1: you could add error handling by adding type safety to check if `user` and `user.roles` is defined.

src/sprints/sprints.controller.ts
```diff
@@ -56,16 +57,28 @@ export class SprintsController {
...
+        summary:
+            "[Roles:Admin] gets all the voyages and sprints details in the database",
     })
     @ApiResponse({
         status: HttpStatus.OK,
         description: "successfully gets all voyage and sprints data",
         type: VoyageResponseWithoutMeetings,
         isArray: true,
     })
+    @ApiResponse({
```
timDeHof, line 72: Duplicate `HttpStatus.UNAUTHORIZED` response on line 72 and line 82. I suggest removing one of them.

src/sprints/sprints.service.ts
```diff
@@ -391,11 +403,6 @@ export class SprintsService {
                             `FormId: ${formId} does not exist.`,
                         );
                     }
```
timDeHof, line 405: The error handling in `addMeetingFormResponse` could be improved by adding a default case for unexpected errors.

src/sprints/sprints.service.ts
```diff
@@ -273,91 +280,88 @@ export class SprintsService {
...
-            }
-        }
+    async deleteMeetingAgenda(agendaId: number, req: CustomRequest) {
+        await manageOwnTeamMeetingOrAgendaById({ user: req.user, agendaId });
+
+        return await this.prisma.agenda.delete({
+            where: {
+                id: agendaId,
+            },
+        });
```
timDeHof, line 343: redundant `await` when return ing the result of `this.prisma.aganda.delete`
```suggestion
        return this.prisma.agenda.delete({
            where: {
                id: agendaId,
            },
        });
```

timDeHof (changes requested):
I have added some comments to code.

src/ability/ability.factory/ability.factory.ts
Ajen07: Are you suggesting to create a global function to check for voyageTeam's IDs and voyageTeamMember's IDs?

cherylli: I don't think user.voyageTeams would be large, it's rare that anyone does more than 10 voyages
It's always good to reduce the number of operations especially database queries, unless it reduce readability
I'm not sure if creating a helper function would help in this case because they would still be called multiple times with a helper function but maybe creating a variable from the mapped values and reuse it might be a good idea

Ajen07: Like creating a variable for mapped values in that ability file only?

timDeHof: that's what I was thinking with the suggestion. Instead of repeating mapping over the same values multiple times, just create variables for them and reuse it.

src/ability/conditions/meetingOrAgenda.ability.ts
```diff
@@ -54,13 +55,14 @@ export const manageOwnTeamMeetingOrAgendaById = async ({
         meetingOrAgendaTeamId = agenda.teamMeeting.voyageTeamId;
     }
 
-    if (user.roles?.includes("admin")) return;
+    if (!user || !user?.roles) {
```
cherylli, line 59: I think this should be if user's role is not voyager or admin

cherylli (commented):
We should use more descriptive commit message rather than `minor changes`

src/ability/conditions/meetingOrAgenda.ability.ts
```diff
@@ -0,0 +1,71 @@
...
+        meetingOrAgendaTeamId = agenda.teamMeeting.voyageTeamId;
+    }
+
+    if (!user.roles.includes("voyager") && !user.roles.includes("admin")) {
+        throw new ForbiddenException(
+            "Invalid user role for Sprint Meeting access control",
+        );
+    }
+
+    if (user.roles.includes("admin")) return;
```
cherylli, line 64: I think this can be moved to above line 58, then just check for voyages in line 58.

timDeHof (approved):
looks good, everything passes in the tests and on swagger.

timDeHof (approved):
looks good, everything passes in the tests and on swagger.

PR #194, opened 2024-08-31
Feature/discord oauth p2 - functionality

cherylli:
# Description

Add discord login functionality, go to http://localhost:8000/api/v1/auth/discord/login, a discord login / authorize screen will pop up, click authorize and it will grant access token, refresh token and then redirect to the frontend url

It gets the user discord email, 
- if this email exists in our database, we add a discord oauth profile to that user
- if the email does not exist, we create a new user + an oauth profile

Currently, if the user has different emails for credentials, and differnet oauth, it will create different accounts as we have no way to match users with different emails. In future, we should make it so user can link their social accounts to their existing email, this way, we can link their accounts without needing to match their emails

## Type of change

- [ ] Tests

# How Has This Been Tested?

Login, and check access_token, refresh_token in cookies
Make sure to db seed before testing, discord needs to be in the oauthProvider table otherwise it will generate an error

# Checklist:


prisma/schema/user.prisma
```diff
@@ -65,7 +65,7 @@ model UserRole {
 
 model OAuthProvider {
   id   Int    @id @default(autoincrement())
-  name String
+  name String @unique
```
cherylli, line 68: I have generated a migrate for this in #190, so I won't be doing it here

timDeHof (dismissed):
All unit and e2e test pass and 0authProvider in the database shows discord in the model

cherylli:
@timDeHof need a re-review if you can please, resolved changelog conflict

Ajen07 (approved):
The Discord login flow is working well, and new user is being added to the database correctly.

timDeHof (approved):
Every unit & e2e tests passed. Also able to login into the dashboard using discord

PR #208, opened 2024-09-21
Teams can create own tech categories

JoshuaHinman:
# Description

This PR completes the first part of task 86b1ym631
adds 3 routes for team tech stack categories (create, update, delete)
These are located in the techs section of the API
Now, each team will have it's own instances of the 6 main categories, and can also create their own
The schema for tech stack categories has been updated, along with seed data

## Issue link

Fixes # [86b1ym631](https://app.clickup.com/t/86b1ym631)

## Type of change

- [ ] Tests

# How Has This Been Tested?

Migrate for DB updates and seed:
npx prisma migrate dev

Test routes in swagger
Verify in DB: 66 categories for 11 teams
Verify duplicate category names for the same teamId throw an exception, and are case-insensitive - i.e. "TEST" === "Test"

# Checklist:

- [ ] My changes generate no new warnings


cherylli:
image

haven't looked at the code yet but saw this in the lint and test workflow, we need to make it nullable or set a default value. At this stage of the project I don't think it matters what we do since prod will be wiped after the alpha test or before the next test phase. Probably better to set a default value because we don't want it null. 

But we need to remove the default value after the new schema is pushed/deployed to all databases.

thoughts?

JoshuaHinman:
> thoughts?
I changed the voyageTeamId relation to optional, even though I don't think it should be.  It passed the check - It looks like the other relations are done this way too.

cherylli:
> I changed the voyageTeamId relation to optional, even though I don't think it should be. It passed the check - It looks like the other relations are done this way too.

sounds good. I think some of the relations are nullable because I needed to do that to set up "onDelete" referential actions on team deletion, when I don't want to cascade delete, there are probably better ways we can look into that later. It will probably be a lot more relavent when we are working on the admin dashboard

cherylli:
removed @Ajen07 as a reviewer as he'll be away for 2 weeks

src/techs/techs.controller.ts
```diff
@@ -228,6 +232,133 @@ export class TechsController {
...
+        status: HttpStatus.CONFLICT,
+        description: "Tech stack category already exists for the team",
+        type: ConflictErrorResponse,
+    })
+    @ApiResponse({
+        status: HttpStatus.NOT_FOUND,
+        description: "Invalid tech stack category id",
+        type: NotFoundErrorResponse,
+    })
+    @Patch("teams/techStackCategory")
```
cherylli, line 308: I think we should have the techstackCategoryId in the uri instead of the body as it identifies the resource

src/techs/techs.controller.ts
```diff
@@ -228,6 +232,133 @@ export class TechsController {
...
+        status: HttpStatus.BAD_REQUEST,
+        description: "Bad Request - tech stack category couldn't be deleted",
+        type: BadRequestErrorResponse,
+    })
+    @ApiResponse({
+        status: HttpStatus.NOT_FOUND,
+        description: "Invalid tech stack category id",
+        type: NotFoundErrorResponse,
+    })
+    @Delete("teams/:teamId/techStackCategory/:techStackCategoryId")
```
cherylli, line 349: we shouldn't need teamId here as techSTackCategory is enough to identify the item we need to delete. for checking if user has the permission to do it, their teamId, teamMemberId are in `user.req`

cherylli (changes requested):
image

is teamId same as voyageTeamId? 


I tested on swagger (POST, PATCH, DELETE) they all work fine, ran all tests, still pass. Left a few comments but it's a good starting point for the change.

JoshuaHinman:
>
> is teamId same as voyageTeamId?
> 
Yes that's right - I actually tried removing this param from the controller, but the response sent an error that said "URL shouldn't be empty".  I couldn't find the source of the message.  Should I remove voyageTeamId from the body?

cherylli:
> > is teamId same as voyageTeamId?
> 
> Yes that's right - I actually tried removing this param from the controller, but the response sent an error that said "URL shouldn't be empty". I couldn't find the source of the message. Should I remove voyageTeamId from the body?

yes please, I think the teamId in uri is good

src/techs/techs.service.ts
```diff
@@ -318,6 +320,109 @@ export class TechsService {
...
+        const teamId = createTechStackCategoryDto.voyageTeamId;
+
+        manageOwnVoyageTeamWithIdParam(req.user, teamId);
+
+        //check if category name with teamid aready exists
+        const categoryAlreadyExists =
+            await this.prisma.techStackCategory.findFirst({
+                where: {
+                    voyageTeamId: createTechStackCategoryDto.voyageTeamId,
+                    name: createTechStackCategoryDto.name,
```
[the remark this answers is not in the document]
JoshuaHinman: How would I approach this?  Does Prisma have a way of designating a case insensitive search?  I haven't found one.  Another possibility would be to apply a standard capitalization to the string value before saving it, so they always follow the same format

timDeHof (approved):
After migrating and seeding the database, all tests passed. Along with teams endpoints in swagger

src/techs/dto/create-techstack-category.dto.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    @IsInt()
+    @IsNotEmpty()
+    @ApiProperty({ example: 1 })
+    voyageTeamId: number;
+
+    @IsInt()
+    @IsNotEmpty()
+    @ApiProperty({ example: 4 })
+    voyageTeamMemberId: number;
```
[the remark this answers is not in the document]
JoshuaHinman: The VoyageTeamMemberValidationPipe requires voyageTeamMemberId to be in the body.  However, there are other functions to verify "own team", so I can remove it.

src/techs/dto/update-techstack-category.dto.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    @IsInt()
+    @IsNotEmpty()
+    @ApiProperty({ example: 1 })
+    categoryId: number;
+
+    @IsInt()
+    @IsNotEmpty()
+    @ApiProperty({ example: 1 })
+    voyageTeamId: number;
```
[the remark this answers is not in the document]
JoshuaHinman: voyageTeamId is used here to verify the user's "own team" and then to verify that the category belongs to that team
image

cherylli:
Got message from @JoshuaHinman saying he'll finish the requested changes after his break.

src/techs/dto/update-techstack-category.dto.ts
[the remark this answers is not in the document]
JoshuaHinman: Ok - I have the other revisions completed.  After going over this suggestion, I think voyageTeamId needs to be stay in the DTO or be included in the URI.  This route checks for duplicate categories before posting to db, and it needs the category name and voyageTeamId to identify them.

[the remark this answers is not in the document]
cherylli: I think we can have the categoryId in the uri as it's identifying the resource. 

I think we won't need the teamId to check - check out `src/ability/conditions/resource.ability.ts`, would something like this work?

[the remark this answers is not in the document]
JoshuaHinman: The teamId isn't used for permissions, it's needed to identify duplicate categories.  Because category names aren't unique anymore, they need to be matched by name and teamId

prisma/schema/teamStack.prisma
```diff
@@ -2,8 +2,10 @@
 
 model TechStackCategory {
   id          Int    @id @default(autoincrement())
-  name        String @unique
+  name        String 
   description String
+  voyageTeam         VoyageTeam?         @relation(fields: [voyageTeamId], references: [id], onUpdate: Cascade, onDelete: Cascade)
+  voyageTeamId       Int?
```
cherylli, line 8: will voyageTeamId ever be null?

src/techs/dto/update-techstack-category.dto.ts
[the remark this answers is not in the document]
cherylli: I see. I was thinking we can find the teamId using the categoryId. Then use that, unless they gave the wrong teamId, which I think they would get a 403 anyway...

image

but then it looks like it would add another db query, I don't think it's a big issue to require an extra input, but then every extra input we'll have to do validation, which I think we already have anyway. 

on another note: 

we could also use `@@unique` teamId + CategoryId, then insert without checking since it would just throw an error if the combination is not unique, which would save one db query based on your current implementation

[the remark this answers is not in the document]
cherylli: I also think you'll need to make teamId not nullable, that's why I also asked that question if it needs to be optional. Now I remember it was because we had existing data and we decided to make it nullable to accomodate old data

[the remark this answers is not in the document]
JoshuaHinman: One problem with using `@@unique[ name, teamId]` is that it won't use case-insensitive mode for the string comparison, so we might have to drop that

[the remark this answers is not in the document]
JoshuaHinman: > Do you think citext would work?

Yeah that looks good, and plus it's already installed
Would I use it the same way as in this example?
image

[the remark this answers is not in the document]
cherylli: yes I think so

prisma/schema/teamStack.prisma
JoshuaHinman: No - it won't ever be null.  I made this change because it was failing the build check when I pushed to github, and this was the easiest solution.  I think it said we can also make the DB reset when building

cherylli: yeah at this stage it won't matter, since we are not actually using the prod database. We could re seed it again. But I think in future when we have cases like this (create a new non nullable column with existing null values), which is vey likely, we need to do something like

1. create a nullable newColumn, migrate, fill it with values manually or with a script
2. make newColumn non-nullable, migrate

or 
1. create a non-nullable newColumn with a default value, migrate
2. remove default, migrate 

This will preserve the original data. We will need to do this in 2 stages, so it would require 2 PRs to pass the build checks.

If we want to do this, we can do step 1 in this PR, then after it's merged, we can do another for step 2

cherylli: Or we could just do a PR now which would include the first step - so we can setup uniqueness in this PR (with step 2)

JoshuaHinman: So - since the column is already nullable , we can fill null values manually or set a default value and migrate?

cherylli: hmm you can try but I thought the migration was getting an error, thats why we have to do a 2 step migration

JoshuaHinman: Ok - I added a voyageTeamIdNew column to the tech stack category schema - no default, and a relation for it

timDeHof (approved):
All unit and e2e tests passed. db shows 66 categories for 11 teams. Also the POST techCategory endpoint throw an exception and is case-sensitive for duplicate category names on the same team id.

CHANGELOG.md
```diff
@@ -60,6 +60,8 @@
 - Add CASL permissions for Team Sprint endpoint ([#193](https://github.com/chingu-x/chingu-dashboard-be/pull/193))
 - Add units tests for the teams resource controller & services([#201](https://github.com/chingu-x/chingu-dashboard-be/pull/201))
 - Add github workflow for PR reminders ([#202](https://github.com/chingu-x/chingu-dashboard-be/pull/202))
+- Added routes for teams to create own tech stack categories([#208](https://github.com/chingu-x/chingu-dashboard-be/pull/208))
```
cherylli, line 63: could you please put this under unreleased, since we I just did a release a few days ago sorry

prisma/schema/voyage.prisma
```diff
@@ -91,6 +91,10 @@ model VoyageTeam {
 
   voyageTeamMembers         VoyageTeamMember[]
   teamTechStackItems        TeamTechStackItem[]
+  techStackCategory         TechStackCategory[] @relation("Team")
+  techStackCategoryNew         TechStackCategory[] @relation("NewTeam")
```
cherylli, line 95: Is "NewTeam" a duplicate?

prisma/schema/teamStack.prisma
cherylli, line 1: is it possible to format this file? I mean just the indentation to make it more readable

prisma/schema/teamStack.prisma
```diff
@@ -2,13 +2,20 @@
 
 model TechStackCategory {
   id          Int    @id @default(autoincrement())
-  name        String @unique
+  name               String             @db.Citext
   description String
+  voyageTeam         VoyageTeam?         @relation("Team", fields: [voyageTeamId], references: [id], onUpdate: Cascade, onDelete: Cascade)
+  voyageTeamId       Int?
+
+  voyageTeamNew         VoyageTeam?         @relation("NewTeam", fields: [voyageTeamIdNew], references: [id], onUpdate: Cascade, onDelete: Cascade)
```
cherylli, line 10: is voyageTeamNew required? I see we have voyageTeam and also voyageTeamNew, voyageTeamNew seems to be unused (all null) in the database

cherylli (changes requested):
Thanks! This is good. I have tested on swagger everything listed in the description works, I just have some questions and very minor changes requested

I assume tests would be added/updated in later parts? Maybe new tests for categories, and update old tests if needed

prisma/schema/teamStack.prisma
JoshuaHinman: I added voyageTeamNew because of our conversation about migrating the nullable column to a new one and then making it non-nullable - yes it is unused right now

cherylli: can it be removed now? or do we need to do it in another PR? would an additional migration work? Is this a temporary column? I think I'm sligtly confused now

JoshuaHinman: Last week you were describing a 2 step migration, starting with making a new column to copy values to.  I thought you wanted to do that in this PR.

JoshuaHinman: From [this](https://github.com/chingu-x/chingu-dashboard-be/pull/208/#discussion_r1805583201) comment

cherylli: If you can sure, sorry I haven't tried or tested anything, we might be able to do multiple migrations here like you have already did (I see 3 migrations here) if not, then we can do it in another PR, we don't want duplicated columns.

I don't think we can do it in 1 PR but I could be wrong, reason being you'll need to migrate it to add `voyageTeamId`, but we cannot make it non-nullable the first migration because it's a new column and old rows has no value, so making it nullable is the only option for this migration to success (or set a default value might work), then we have to patch null values to non-null on prod server then migrate again to update column definition to make `voyageTeamId` non-nullable

I see both voyageTeamId and newVoyageTeamId are still nullable, but you seem to be able to set a composite unique, so I think we don't need a temp column now, and we can migrate, patch the existing data, then make it non optional? 

Intially, I thought you won't be able to add `@@unique` due to voyageTeamId being optional, in this case it would be necessary to create a non-nullable newVoyageTeamId, use that instead and rename it later on.

But now I think this method would be more straightforward 
image

I hope this explains the situation better

JoshuaHinman: Yeah sounds good - I'll remove the new column

JoshuaHinman: Ok - I've removed the new column from the categories schema , cleaned up the migrations folder , updated changelog.  Should be good to go

cherylli (approved):
Just randomly discovered this bug, but I think this may not be in the scope of this PR?

image

voyageTeam 1 doesn't own category 51, and I logged in as dan who is in voyageTeam1 but not in the team which owns category 51 (team 9)


==== 

Other things i've manually tested and all worked
- create a new tech stack category, throw 409 when duplicated (non case sensitive) as expected 
- added the same category to different teams
- try to add category to someone elses team (403)
- patch (own team and other teams)
- delete (own team and other teams)

JoshuaHinman:
> voyageTeam 1 doesn't own category 51, and I logged in as dan who is in voyageTeam1 but not in the team which owns category 51 (team 9)
> 
I don't think it's specific to this PR but it could be related.
I've seen similar issues before , I think it was because  the user was an admin.  It looks like Dan is an admin - is that causing the problem?

cherylli:
> > voyageTeam 1 doesn't own category 51, and I logged in as dan who is in voyageTeam1 but not in the team which owns category 51 (team 9)
> 
> I don't think it's specific to this PR but it could be related. I've seen similar issues before , I think it was because the user was an admin. It looks like Dan is an admin - is that causing the problem?

In my database dan is not an admin, his only role is "voyager" but yeah if you want to fix this in another parts its fine, it's not a huge issue which stops the app from working.

cherylli:
hi @timDeHof are you able to re- review this PR please

timDeHof (approved):
All tests passed! migrating db works. Also the preventing the duplicate tech category name works.

PR #212, opened 2024-10-04
add same site property to cookie

Dan-Y-Ko:
# Description

I'm having trouble getting a screenshot for this but basically there is an issue with the cookies and cross-site domains so we need to set the same-site as none. Adding that fixed this issue

This is probably only an issue with client components but since we are gunna have a separate react project in the future (which would be client) that'll communicate with this backend, this would fix for that usecase as well

## Issue link

Fixes # (issue)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Tests

# How Has This Been Tested?

I ran into this issue when trying to test from localhost on frontend and the deployed backend dev server.
I also ran into this issue using localhost on the backend and the preview deployed url on vercel (it's using the localhost backend, just different domain).

After implementing this, it worked as expected

# Checklist:

### DIFF
--- a/src/auth/auth.service.ts
+++ b/src/auth/auth.service.ts
@@ -134,11 +134,13 @@ export class AuthService {
             maxAge: AT_MAX_AGE * 1000,
             httpOnly: true,
             secure: true,
+            sameSite: "none",
         });
         res.cookie("refresh_token", refresh_token, {
             maxAge: RT_MAX_AGE * 1000,
             httpOnly: true,
             secure: true,
+            sameSite: "none",
         });
     }
 
--- a/src/auth/auth.service.ts
+++ b/src/auth/auth.service.ts
@@ -134,7 +134,7 @@ export class AuthService {
             maxAge: AT_MAX_AGE * 1000,
             httpOnly: true,
             secure: true,
-            sameSite: "none",
+            sameSite: "none", // TODO: remove this in future when we use actual domains for the dashboard, they will likely be same-site
         });
         res.cookie("refresh_token", refresh_token, {
             maxAge: RT_MAX_AGE * 1000,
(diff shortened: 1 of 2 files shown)


src/auth/auth.service.ts
```diff
@@ -134,11 +134,13 @@ export class AuthService {
             maxAge: AT_MAX_AGE * 1000,
             httpOnly: true,
             secure: true,
+            sameSite: "none",
```
cherylli, line 137:
```suggestion
            sameSite: "none", // TODO: remove this in future when we use actual domains for the dashboard, they will likely be same-site
```

cherylli: just want to add a note here so we don't forget
I imagine the dashboard could be something like dashboard.chingu.io, and backend would be api.chingu.io

cherylli (dismissed):

Dan-Y-Ko:
I do want to point out that if you remove this later, we won't be able to use the deployed be dev server and we will have to use the local dev server. I don't really want our team to be forced to do that tbh

cherylli:
> I do want to point out that if you remove this later, we won't be able to use the deployed be dev server and we will have to use the local dev server. I don't really want our team to be forced to do that tbh

you are right, maybe we need to have a condition statement for dev/prod environment.

src/auth/auth.service.ts
```diff
@@ -134,11 +134,13 @@ export class AuthService {
             maxAge: AT_MAX_AGE * 1000,
             httpOnly: true,
             secure: true,
+            sameSite: "none", // TODO: remove this in future when we use actual domains for the dashboard, they will likely be same-site
```
cherylli, line 137:
```suggestion
            sameSite: "none",
```

cherylli: remove this for now and put on click up instead, I need to do more research on how it is usually handled. 

I'm thinking set cookies based on the environment, i.e. NODE_ENV

cherylli (approved):

timDeHof (approved):
All tests passed. But I think the `returnTokensOnLoginSuccess` should be wrapped in a try-catch for error handling. Also it should have a explicit return type for clarity and good practices.

PR #221, opened 2024-11-08
Update tech selections

JoshuaHinman:
# Description

- This PR revises the PATCH /teams/techs/:techId route to only update a single tech stack selection, setting the `isSelected` value
- Also, a permission bug has been fixed for POST /teams/{teamId}/techs , ensuring users can only add a new tech for thier own team

## Issue link

Fixes # [86b1ym631](https://app.clickup.com/t/86b1ym631)

## Type of change

- [x] Tests

# How Has This Been Tested?

- migrate db for updates
- yarn test

# Checklist:

### DIFF
--- a/src/techs/techs.controller.ts
+++ b/src/techs/techs.controller.ts
@@ -477,16 +477,16 @@ export class TechsController {
         type: ForbiddenErrorResponse,
     })
     @CheckAbilities({ action: Action.Update, subject: "TeamTechStackItem" })
-    @Patch("teams/techs/:techId")
+    @Patch("techs/:techId/selection")
     updateTechStackSelections(
         @Request() req,
         @Param("techId", ParseIntPipe) techId: number,
-        @Body(ValidationPipe) techSelectionDto: UpdateTechSelectionDto,
+        @Body(ValidationPipe) upateTechSelectionDto: UpdateTechSelectionDto,
     ) {
         return this.techsService.updateTechStackSelections(
             req,
             techId,
-            techSelectionDto,
+            upateTechSelectionDto,
         );
     }
 }
--- a/src/techs/techs.controller.ts
+++ b/src/techs/techs.controller.ts
@@ -476,13 +476,6 @@ export class TechsController {
         description: "forbidden - user does not have the required permission",
         type: ForbiddenErrorResponse,
     })
-    @ApiParam({
-        name: "teamId",
-        description: "voyage team Id",
-        type: "Integer",
-        required: true,
-        example: 2,
-    })
     @CheckAbilities({ action: Action.Update, subject: "TeamTechStackItem" })
     @Patch("teams/techs/:techId")
     updateTechStackSelections(
--- a/src/techs/techs.controller.ts
+++ b/src/techs/techs.controller.ts
@@ -14,7 +14,7 @@ import {
 import { TechsService } from "./techs.service";
 import { ApiOperation, ApiParam, ApiResponse, ApiTags } from "@nestjs/swagger";
 import { CreateTeamTechDto } from "./dto/create-tech.dto";
-import { TechCategoryDto } from "./dto/update-tech-selections.dto";
+import { UpdateTechSelectionsDto } from "./dto/update-tech-selections.dto";
 import { CreateTechStackCategoryDto } from "./dto/create-techstack-category.dto";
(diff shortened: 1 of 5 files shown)


src/techs/techs.controller.ts
```diff
@@ -477,24 +476,17 @@ export class TechsController {
...
-    })
     @CheckAbilities({ action: Action.Update, subject: "TeamTechStackItem" })
-    @Patch("teams/:teamId/techs/selections")
+    @Patch("teams/techs/:techId")
     updateTechStackSelections(
         @Request() req,
-        @Param("teamId", ParseIntPipe) teamId: number,
-        @Body(ValidationPipe) updateTechSelectionsDto: UpdateTechSelectionsDto,
+        @Param("techId", ParseIntPipe) techId: number,
+        @Body(ValidationPipe) techSelectionDto: UpdateTechSelectionDto,
```
cherylli, line 484: should we leave it as updateTechSelectionDto just to be more clear that we are using the UpdateTechSelectionDto? And also consistent with the function `techs.service.ts`

src/techs/techs.service.ts
```diff
@@ -74,46 +74,53 @@ export class TechsService {
...
+                id: techId,
+            },
+            select: {
+                categoryId: true,
+                voyageTeamId: true,
+            },
+        });
+        if (!tech) {
+            throw new NotFoundException(
+                `Tech ${MAX_SELECTION_COUNT} not found.`,
```
cherylli, line 92: is this supposed to be techId instead of `MAX_SELECTION_COUNT`?

src/techs/techs.service.ts
```diff
@@ -124,8 +131,14 @@ export class TechsService {
         //check for valid teamId
         await this.validateTeamId(teamId);
 
+        //check if user is a member of the team
         manageOwnVoyageTeamWithIdParam(req.user, teamId);
```
cherylli, line 135: I think we won't need `manageOwnVoyageTeamWithIdParam` if we have `userCanChangeCategory`
Also could we move `userCanChangeCategory` to the `ability` folder?

src/techs/techs.service.ts
```diff
@@ -353,7 +366,8 @@ export class TechsService {
         techStackCategoryId: number,
         updateTechStackCategoryDto: UpdateTechStackCategoryDto,
     ) {
-        const permission = await this.userCanChangeCategory(
+        const permission = true;
```
cherylli, line 369: how come permission is set to true but then a few lines below there is code to throw an error when permission is not truthy?

test/techs.e2e-spec.ts
```diff
@@ -677,142 +677,113 @@ describe("Techs Controller (e2e)", () => {
         });
     });
 
-    describe("PATCH voyages/teams/:teamId/techs/selections - updates isSelected value of tech stack items", () => {
-        it("should return 200 and an array of updated techs, if successful", async () => {
-            const teamId: number = 2;
+    describe("PATCH voyages/teams/:teamId/techs/selections - updates isSelected value of a tech stack tems", () => {
```
cherylli, line 680:
```suggestion
    describe("PATCH voyages/teams/:teamId/techs/selections - updates isSelected value of a tech stack item", () => {
```

src/techs/techs.service.ts
JoshuaHinman: oops - I did this as a shortcut when refactoring and forgot about it...

src/techs/techs.service.ts
JoshuaHinman: I tried removing this, but got a testing error.  Looking at it closer, these functions are comparing different values.  `manageOwnVoyageTeamWithIdParam` verifies that user is in the team from the URI params, and `userCanChangeCategory` verifies that the user is in the team from `category.voyageTeamId`

cherylli: but I mean if the user is in the team in `category.voyageTeamId` then we can assume they are allowed to update it, without having to check if the user is in the team from the uri params? I guess we only need to check the one which is more restrictive

JoshuaHinman: Ok - took it out and fixed the test

curtwl (approved):
lgtm

cherylli (approved):
Did some testing on swagger and ran all the tests, working as expected

PR #222, opened 2024-11-12
Feature/oauth GitHub

curtwl:
# Description

### To test this PR
1. obtain github oauth client id and client secret (on clickup, you can also make your own for dev), and put them in .env
2. `npx prisma db push` for both dev and test db
3. go to `http://localhost:8000/api/v1/auth/github/login` a github authorization table will pop up, and it should return the user's basic github profile 

## Issue link

Fixes # (issue)

## Type of change

- [ ] Tests
- [x] This change requires a documentation update

# How Has This Been Tested?

Manually and automated tests

# Checklist:

- [ x] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] I have made corresponding changes to the documentation
- [ x] My changes generate no new warnings
- [ x] I have added tests that prove my fix is effective or that my feature works
- [ x] New and existing tests pass locally with my changes
- [ x] Any dependent changes have been merged and published in downstream modules
- [ x] I have updated the change log


cherylli (changes requested):
image

I suggest adding a more desciptive error message for this error: upsert error because there's no github in the provider database

We need to update `prisma/seed/data/oauth-providers.ts` to include github. We also need to delete the file `prisma/seed/oauth.ts` as it's not used (mistake from a previous PR)

everything else work fine

cherylli:
image

image

invalid code is not handled - I'm quite sure the discord one also has the same issue

If you can fix both it would be great, if not I can fix the discord one in another PR

curtwl:
> invalid code is not handled - I'm quite sure the discord one also has the same issue
> 
> If you can fix both it would be great, if not I can fix the discord one in another PR

Does `{"message":"Failed to obtain access token. Possibly because of invalid redirect code.","error":"Bad Request","statusCode":400}` work for the error message? I guess other situations could cause the same message since it's just a generic error for the super.canActivate(context) in the AuthGuards.

cherylli:
> {"message":"Failed to obtain access token. Possibly because of invalid redirect code.","error":"Bad Request","statusCode":400}

yes, this sounds good. It is descriptive enough to know the cause. Thanks!

test/auth.e2e-spec.ts
```diff
@@ -683,4 +683,24 @@ describe("AuthController e2e Tests", () => {
...
         });
     });
+
+    describe("Initiate Github OAuth GET /auth/github/login", () => {
+        it("should redirect ", async () => {
+            const res = await request(app.getHttpServer())
+                .get("/auth/github/login")
+                .expect(302);
+
+            const clientId = Oauth.github.clientId;
```
andram11, line 693: I see that in the beforeAll() part of the test suite, we declare a provider for discord but i don't see any for Github. The github part is declared indeed in the Oauth variable but shouldn't it also be added as a new useValue?

test/auth.e2e-spec.ts
andram11, line 1: Probably not related but I am getting a test failed for this suite, specifically in this test (running via Docket); AuthController e2e Tests › Request password reset POST /auth/reset-password/request › should return 200 if user account exist, resetToken should be in the database
Instead i get this: {
    "statusCode": 401,
    "message": "Unsuccessful: Status Code: \"401\" Message: \"API key authentication/authorization failure. You may be unauthorized to access the API or your API key may be expired. Visit API keys management section to check your keys.\""
}

src/auth/github-auth.service.ts
```diff
@@ -0,0 +1,99 @@
...
+                                },
+                            },
+                            providerUserId: user.githubId,
+                            providerUsername: user.username,
+                        },
+                    },
+                },
+            });
+        } catch (e) {
+            if (e.code === "P2025") {
```
andram11, line 87: What happens if the code is not P2025 (PrismaClientKnownRequestError)? I don't see an error being thrown for any other kind of errors in this catch statement

cherylli: if it's not a P2025 error it would just re-throw the error which is handled in the `prisma-client-exception.filter.ts` Exception filter if it's a prisma exception. if not it would just be catched by the default nest.js exception filter https://docs.nestjs.com/exception-filters

test/auth.e2e-spec.ts
cherylli: You'll need to run the tests inside docker, using the `test:docker` command
Also make sure the test database is "prepared". I.e. push the schema

curtwl:
> invalid code is not handled - I'm quite sure the discord one also has the same issue
> 
> If you can fix both it would be great, if not I can fix the discord one in another PR

Strangely enough, Discord gives a specific invalid code error instead of the generic one Github gives. I *think* it might be a bug in the passport-github2 package, but I'll keep poking around while working on LinkedIn to see if there's a way to specifcally catch invalid codes for Github too.

cherylli:
> image
> 
> image
> 
> invalid code is not handled - I'm quite sure the discord one also has the same issue
> 
> If you can fix both it would be great, if not I can fix the discord one in another PR

This is working for github but not discord, that's because discord message is actually `Invalid "code"`, not "Invalid code"

image


I suggest matching it with the error code `invalid_grant` instead of the error message as it is more likely to be changed by the oauth providers, like so 
image

cherylli:
> a specific invalid code error instead of the generic one

ah I posted just a few mins after you posted, so I didn't see that but I've posted why it wasn't working. I changed it to `Invalid "code"` and it worked, but yeah I think using e.code is probably a better option

curtwl:
> > a specific invalid code error instead of the generic one
> 
> ah I posted just a few mins after you posted, so I didn't see that but I've posted why it wasn't working. I changed it to `Invalid "code"` and it worked, but yeah I think using e.code is probably a better option

Yeah, I tried to match invalid_grant with Github but even looking through the entire response, there was nothing specific like that.

cherylli:
> > > a specific invalid code error instead of the generic one
> > 
> > 
> > ah I posted just a few mins after you posted, so I didn't see that but I've posted why it wasn't working. I changed it to `Invalid "code"` and it worked, but yeah I think using e.code is probably a better option
> 
> Yeah, I tried to match invalid_grant with Github but even looking through the entire response, there was nothing specific like that.

They might have a different code, I didn't check the github one, but try to print `e` and you might see it

curtwl:
> > Yeah, I tried to match invalid_grant with Github but even looking through the entire response, there was nothing specific like that.
> 
> They might have a different code, I didn't check the github one, but try to print `e` and you might see it

That's what I meant, I printed e and looked for both invalid_grant and anything specific at all but it was just all non-specific.

cherylli:
> > > Yeah, I tried to match invalid_grant with Github but even looking through the entire response, there was nothing specific like that.
> > 
> > 
> > They might have a different code, I didn't check the github one, but try to print `e` and you might see it
> 
> That's what I meant, I printed e and looked for both invalid_grant and anything specific at all but it was just all non-specific.

In that case I think what you have for github is fine, I normally just relay the error message if there's no specific error.

andram11 (dismissed):

cherylli:
sorry totally missed this new push 2 days ago
image

src/auth/strategies/github.strategy.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    async validate(
+        accessToken: string,
+        refreshToken: string,
+        profile: Profile,
+    ): Promise<any> {
+        const { username, id, photos, emails } = profile;
+
+        const avatar = photos && photos.length > 0 ? photos[0].value : null;
+        const email = emails && emails.length > 0 ? emails[0].value : null;
```
cherylli, line 32: I wonder if we should match/check all the (github verified) emails with existing users in the database, instead of just the first email. 
If none of them is found in the database we can create a new account with the primary email

curtwl: Does discord need this functionality too?

cherylli: probably, but I don't think discord lets you have multiple emails

cherylli:
@curtwl do you feel like squashing the commits? This might be a good practice if you want to try it out. Otherwise we can probably just leave it, next release notes will probably be manually updated combining our changelog and release please change log

cherylli:
One more question - do you know if oauth only returns verified emails? or is there a way to check if the emails are verified? 

Otherwise if users add emails which do not belong to them we don't want to automatically link to an existing account

src/auth/strategies/github.strategy.ts
```diff
@@ -0,0 +1,46 @@
...
+            scope: ["identify", "user:email"],
+        });
+    }
+
+    async validate(
+        accessToken: string,
+        refreshToken: string,
+        profile: GithubProfile,
+    ): Promise<any> {
+        const { username, id, photos, emails } = profile;
```
timDeHof, line 30: I'm getting an error saying that "'username, id, photos' properties does not exist on type `GitHubProfile`". Also I found a types package for `passport-github2` at [@types/passport-github2](https://classic.yarnpkg.com/en/package/@types/passport-github2) on yarn.

curtwl:
> @curtwl do you feel like squashing the commits? This might be a good practice if you want to try it out. Otherwise we can probably just leave it, next release notes will probably be manually updated combining our changelog and release please change log

Sure, I've been reading through the docs and it doesn't seem like it'll take long after I understand the workflow.

src/auth/strategies/github.strategy.ts
curtwl: I think the types package got lost resolving a merge conflict, oops! Definitely had that installed while developing.

timDeHof: ok, I'll look at again once you install the types package.

src/auth/discord-auth.service.ts
```diff
@@ -78,4 +80,19 @@ export class DiscordAuthService implements IAuthProvider {
             },
         });
     }
+
+    async findUserByEmails(emails): Promise<any | null> {
```
davideastmond, line 84: Typescript is complaining. 
```suggestion
    async findUserByEmails(emails: string): Promise<any | null> {
```

davideastmond, line 84: I'm wondering if we can get a better return type for this method? Can we actually return a properly typed user?

If we're going to use `any` as a return type here, you shouldn't need the union with `null`, since any should cover the `null` type.

src/auth/discord-auth.service.ts
```diff
@@ -40,10 +40,12 @@ export class DiscordAuthService implements IAuthProvider {
                 "[discord-auth.service]: Cannot get email from discord to create a new Chingu account",
             );
 
+        const existingUser = await this.findUserByEmails([user.email]);
```
davideastmond, line 43: Is it possible that this could return null?

src/auth/github-auth.service.ts
```diff
@@ -0,0 +1,121 @@
...
+import { generatePasswordHash } from "../global/auth/utils";
+import { OAuthConfig } from "@/config/Oauth/oauthConfig.interface";
+
+@Injectable()
+export class GithubAuthService implements IAuthProvider {
+    constructor(
+        private prisma: PrismaService,
+        @Inject("OAuth-Config") private oAuthConfig: OAuthConfig,
+    ) {}
+    async validateUser(user: GithubUser) {
```
davideastmond, line 22: Just a tiny suggestion to make the intelliSense clearer, since Typescript was inferring unnecessarily a union type of objects with the same properties 
```suggestion
    async validateUser(user: GithubUser): Promise<AuthUserResult> {
```

src/auth/github-auth.service.ts
```diff
@@ -0,0 +1,121 @@
...
+        // and will be able to login with this in future,
+        // or maybe the app will prompt user the input the password, exact oauth flow is to be determined
+
+        // this should not happen when "user:email" is in the scope
+        if (!user.email)
+            throw new InternalServerErrorException(
+                "[github-auth.service]: Cannot get email from github to create a new Chingu account",
+            );
+
+        const existingUser = await this.findUserByEmails([
```
davideastmond, line 47: Is it possible that this could return `null`?

src/auth/guards/github-auth.guard.ts
```diff
@@ -0,0 +1,32 @@
...
+                    `Failed to obtain access token. Possibly because of invalid redirect code.`,
+                );
+            }
+            throw e;
+        }
+        const request = context.switchToHttp().getRequest();
+        await super.logIn(request);
+
+        return activate;
+    }
```
davideastmond, line 31: Similar comment here that I made in the file `src/auth/guards/discord-auth.guard.ts`

src/auth/strategies/github.strategy.ts
```diff
@@ -0,0 +1,46 @@
...
+            clientID: clientId,
+            clientSecret: clientSecret,
+            callbackURL: callbackUrl,
+            scope: ["identify", "user:email"],
+        });
+    }
+
+    async validate(
+        accessToken: string,
+        refreshToken: string,
```
davideastmond, line 27: It doesn't look like we're using these variables?

src/auth/strategies/github.strategy.ts
```diff
@@ -0,0 +1,46 @@
...
+        profile: GithubProfile,
+    ): Promise<any> {
+        const { username, id, photos, emails } = profile;
+        const avatar = photos && photos.length > 0 ? photos[0].value : null;
```
davideastmond, line 31: Might be related to the comment above about the types packages - for readability I think we should explicitly declare what the type of avatar is

src/auth/discord-auth.service.ts
cherylli: Currently, this is a prisma User Model type, I think there's one we could use, for reference: https://www.prisma.io/docs/orm/prisma-client/type-safety. The current implicit type is this

image

src/auth/discord-auth.service.ts
cherylli: Yes it can https://www.prisma.io/docs/orm/reference/prisma-client-reference#findunique

src/auth/github-auth.service.ts
cherylli: Is AuthUserResult something built in or something we have to define?
I think this might be same as the prisma User model type

src/auth/github-auth.service.ts
cherylli: yes according to the Prisma API https://www.prisma.io/docs/orm/reference/prisma-client-reference#findfirst

PR #230, opened 2024-12-07
Fixed server crash bug when tokens not found in db

andram11:
# Description

Fix for bug: cannot read properties of undefined when user is logged in but database re-seeded, and user is still logged in (still has access and refresh tokens) and tries to access non public endpoints.

## Issue link

Fixes  [bug](https://app.clickup.com/t/86b2k9ua1)

## Type of change

- [ ] Tests

# How Has This Been Tested?

Before fix, I log in with a user and have an access & refresh token.
If I shut down server and reseed the database, then I try to access a non public endpoint (like get forms by id) using the access & refresh token, I received an http 500 Internal Server Error. In logs there is a 'cannot read properties of undefined'(reading 'map')...After some time, when the token expires, I get the usual 403 error. This error occurs because when treating the JWT token, we were retrieving the roles of a user even though the user id no longer existed in the db (after reseed). As such, the roles were undefined which caused issues when calculating the abilities of user. 
After fix, If i reseed and use old access & refresh token I just get a 403.

# Checklist:


src/auth/strategies/at.strategy.ts
```diff
@@ -34,20 +34,28 @@ export class AtStrategy extends PassportStrategy(Strategy, "jwt-at") {
     }
 
     async validate(payload: any) {
-        const userInDb = await this.usersService.getUserRolesById(payload.sub);
+        //Check if user id provided in payload exists in db
+        const userInDb = await this.usersService.findUserById(payload.sub);
```
cherylli, line 38: The problem with this approach is this is another extra db query on a semi frequent operation.

Is it possible to use the original query but throw an error if `!userInDb`? I think it returns null when nothing is found

https://www.prisma.io/docs/orm/reference/prisma-client-reference#findunique

andram11: The original query uses only getUserRolesById which will return only the formatted user roles even when it finds no user in the db: it will just return {roles:undefined} for example. We could refactor getUserRolesById to only return the roles when the user actually exists, that would indeed avoid the extra query

cherylli: yes I think that's a better option, we won't need to return `roles:undefined` if there's no user. And yeah that makes sense I guess because we have a select, so it would return those instead of a null on a basic findUnique

cherylli (commented):
Tested manually and ran all the tests, all good.
I'm glad you figured out how to reproduce the error I've completed forgotten to put that in and I've also forgotten how to trigger the error to test it 😂

Just need a small change on the changelog, since this didn't get merged before the release.

JoshuaHinman (approved):
All tests passed and swagger tests worked as described - no 500s

cherylli (dismissed):
thanks!

JoshuaHinman (approved):
All good!

cherylli:
@andram11 could you please resolve merge conflicts when you have a chance? Thanks!

cherylli (dismissed):

cherylli:
hey @andram11, I noticed from the last couple of PR the changelog entries were in the wrong place, when you resolve conflict could you update it like below? Thanks

https://github.com/chingu-x/chingu-dashboard-be/pull/235/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

JoshuaHinman (dismissed):
Changes look good

cherylli (dismissed):
re-approve after fixing changelog

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@
 - Updated readme for installation part ([#225])(https://github.com/chingu-x/chingu-dashboard-be/pull/225)
 - Updated nestjs packages to latest version ([#233])(https://github.com/chingu-x/chingu-dashboard-be/pull/233)
 - Refactoring of email service + unit tests ([#232](https://github.com/chingu-x/chingu-dashboard-be/pull/232))
+- Added version release link to Swagger docs ([#218](https://github.com/chingu-x/chingu-dashboard-be/pull/231))
```
cherylli, line 26: this line is duplicated

cherylli (dismissed):

cherylli (approved):
re-approve again after resolving conflicts

cherylli:
remove all changes to the changelog, since we've stopped manually updating changing

cherylli (approved):

src/auth/strategies/at.strategy.ts
```diff
@@ -34,20 +34,24 @@ export class AtStrategy extends PassportStrategy(Strategy, "jwt-at") {
     }
 
     async validate(payload: any) {
-        const userInDb = await this.usersService.getUserRolesById(payload.sub);
+        //Get user roles
```
davideastmond, line 37: Would it be helpful to annotate the return type of this validate method? It looks like it would be Promise<UserReq | undefined>

jenny-alexander (approved):
I was able to replicate issue and get a 500 response. 
After checking out these changes, I ran 'npm run seed' and accessed GET endpoint /api/v1/teams, received 401 error.

@cherylli - I get 401 (unauthorized) but the PR description explicitly states getting 403 (forbidden). I approved the PR anyway since this might be a typo in the PR text itself?

cherylli:
> I was able to replicate issue and get a 500 response. After checking out these changes, I ran 'npm run seed' and accessed GET endpoint /api/v1/teams, received 401 error.
> 
> @cherylli - I get 401 (unauthorized) but the PR description explicitly states getting 403 (forbidden). I approved the PR anyway since this might be a typo in the PR text itself?

yeah 401 makes more sense since the user is not logged in. I think I got 401 too, then the user will be required to relogin in this case

src/auth/strategies/at.strategy.ts
cherylli: Good idea, updated it

davideastmond (approved):

jenny-alexander (approved):

PR #231, opened 2024-12-12
Updated Swagger version link to include latest release link and latest tag

andram11:
# Description

Not sure this is what was requested by the story but:
- i added an env variable LATEST_RELEASE with the link to the latest github release
- i adapted app config & service to use this env variable
- i read the link from the env and extract the tag version
- i added the "latest-dev" tag to indicate it's always the latest release

## Issue link

Fixes # (https://app.clickup.com/t/86b32rm4d)

## Type of change

- [ ] Tests

# How Has This Been Tested?

Manually by checking the docs endpoint

# Checklist:


JoshuaHinman (dismissed):
It took me a few minutes to figure out what uri to use for `LATEST_RELEASE` ,  but I got it working and it displays the version and link in swagger

timDeHof:
I'm not sure what is request for the user story too. If its the LATEST_RELEASE of the dashboard API, then where do I find it on the git repo? Why couldn't it be added to the `package.json` instead of with the environment variables? If its the swagger UI then I found this page in the [docs](https://swagger.io/docs/open-source-tools/swagger-ui/usage/version-detection/)

cherylli:
I also think it would be a better idea to keep track and pull the version from package.json if possible. We have 4 environments now it would be a pain to have to update all of them for each release

this could be useful https://betterstack.com/community/questions/how-to-read-package-version-in-nodejs-code/

src/main.ts
```diff
@@ -28,11 +28,12 @@ async function bootstrap() {
...
     const NODE_ENV = app.get(AppConfigService).nodeEnv;
+    const LATEST_RELEASE = app.get(AppConfigService).latestReleaseLink;
     if (NODE_ENV !== "production") {
         const config = new DocumentBuilder()
             .setTitle("Chingu Dashboard Project")
             .setDescription(
-                "Chingu Dashboard API<br> default access: logged in (user)",
+                `Chingu Dashboard API <a href='${LATEST_RELEASE}'> Version: ${LATEST_RELEASE.split("tag/")[1]} (Latest-dev) </a> <br> default access: logged in (user)`,
             )
             .setVersion("1.0")
```
timDeHof, line 38: I think this should be updated too.

timDeHof (approved):
This works with a direct link to the latest github release. But in the future, we may want to automate it.

src/config/app/app.config.ts
```diff
@@ -4,4 +4,5 @@ export default registerAs("app", () => ({
     nodeEnv: process.env.NODE_ENV,
     port: process.env.PORT,
     frontendUrl: process.env.FRONTEND_URL,
+    latestReleaseLink: process.env.LATEST_RELEASE,
```
cherylli, line 7: I would update this variable link to LATEST_RELEASE_LINk so it's clear that we're putting a link here not the version number

src/main.ts
cherylli: it crashed when the env is not set, need to put a `?`

cherylli: I relase the link we have to use is `https://github.com/chingu-x/chingu-dashboard-be/releases/tag/1.1.0-alpha` I thought it would be something more generic like `https://api.github.com/repos/chingu-x/chingu-dashboard-be/releases`

I don't think this will work it would be a pain to update all the env variables in all the deployments for every release.

Also, this is not a "secret" so shouldn't need to be in env. I guess you put it in env because you were thinking to have a different value in each environment?

Originally I was thinking to just hardcode it on every release, but I think a better solution is getting the version from package.json and we'd just make sure it's in sync with the github release version (we can add this step to the release procedure)

Another thing is I also don't want it to show it being the "latest" version for all the deployments.

cherylli:
@andram11 could you please resolve merge conflicts when you have a chance? Thanks!

andram11:
Updated the conflict and merged dev into local branch

cherylli (dismissed):
Works perfectly if I update the version in package.json
image

Is there an automated way to only show `(latest dev)` on the dev deployment (not the "stable" deployment)? 

Otherwise, we could just leave it as it is now, have "latest dev" there, and hardcode to remove it on the stable version (I don't expect to update the stable version very often, at most every 2-3 months)

Prod won't have access to swagger, so it doesn't matter

andram11:
I looked into this: Is there an automated way to only show (latest dev) on the dev deployment (not the "stable" deployment)? but i couldn't really think of something that can be done at the level of code since it's unaware of the environment on which it is deployed. Other than adding an additional env variable which makes the distinction between dev & stable dev, nothing came up as an immediate solution

cherylli:
> I looked into this: Is there an automated way to only show (latest dev) on the dev deployment (not the "stable" deployment)? but i couldn't really think of something that can be done at the level of code since it's unaware of the environment on which it is deployed. Other than adding an additional env variable which makes the distinction between dev & stable dev, nothing came up as an immediate solution

thats fine then, manually changing it is fine for now

cherylli (approved):

timDeHof (approved):
Looks Great! All the tests passed too.
image

PR #238, opened 2025-02-07
feat: Stage 1 - add parseConfig field to Question and OptionChoice models

timDeHof:
# Description

Added parseConfig JSON field to Question and OptionChoice models to support dynamic text parsing functionality. This enhancement allows for configurable text substitutions in form questions and option choices, improving the flexibility of form content management.

Key changes:
- Added parseConfig JSONB column to Question and OptionChoice tables
- Updated seed data to include parseConfig mappings for:
  - Sprint check-in form emojis (🚀)
  - Yes/No response mappings
  - Likelihood scale labels
- Updated related test files to accommodate the new field
## Issue link

Fixes # [clickup task 86b3u7wv2](https://app.clickup.com/t/86b3u7wv2)

## Type of change

- [x] Tests

# How Has This Been Tested?

1. Database Migration Tests:
   - Verified successful migration of parseConfig fields
   - Confirmed existing data integrity

2. Unit Tests:
   - Updated and ran sprints controller and service tests
   - Verified mock data handling of new parseConfig field

3. Seed Data Validation:
   - Confirmed proper parsing of template strings in check-in forms
   - Validated Yes/No mappings in form questions
   - Tested emoji substitutions in sprint status options

# Checklist:


cherylli:
I don't think form needs a parse option (I know it's in the docs, maybe I was thinking something else when I wrote that, sorry)

The parseOption for choices should be in the option choice model as the options are pulled from there, the group itself won't need one (it's used to group all the options for the same question)

timDeHof:
> I don't think form needs a parse option (I know it's in the docs, maybe I was thinking something else when I wrote that, sorry)
> 
> The parseOption for choices should be in the option choice model as the options are pulled from there, the group itself won't need one (it's used to group all the options for the same question)

So we only want it on the `OptionChoice` model in `form.prisma`?

cherylli:
> > I don't think form needs a parse option (I know it's in the docs, maybe I was thinking something else when I wrote that, sorry)
> > The parseOption for choices should be in the option choice model as the options are pulled from there, the group itself won't need one (it's used to group all the options for the same question)
> 
> So we only want it on the `OptionChoice` model in `form.prisma`?

That and question. Anywhere with these things `{{ parseOption }}`
examples

OptionChoices: 
image


Questions:
image


So when frontend pull the questions, they will also pull the parseOption Json field in the same table, they use use that to reconstruct the questions 

like so, 

GET /forms

the response will look like this

```
"optionChoices": [
      {
        "id": 17,
        "text": "We have had a good start!",
        "parse": { icon: "greenRocket" }
      },
```
instead of what we have now,

```
"optionChoices": [
      {
        "id": 17,
        "text": "{{greenRocket}} We have had a good start!"
      },
```

prisma/seed/forms/checkinform.ts
```diff
@@ -74,12 +74,21 @@ export const populateCheckinForm = async () => {
                                         data: [
                                             {
                                                 text: "{{greenRocket}} We have had a good start!",
+                                                parseConfig: {
+                                                    greenRocket: "🚀",
```
cherylli, line 78: I think this needs to be something like 
```js
parseConfig: {
  icon: "greenRocket"
  iconUrl: "....png" // if we want to add this in future, we can, but not required as frontend will match greenRocket with their own icon
}

```

prisma/seed/forms/voyage-project-submission.ts
```diff
@@ -94,6 +94,10 @@ export const populateVoyageSubmissionForm = async () => {
                             },
                         },
                         text: "{{Not Likely,Extremely Likely}}On a scale of 0-10, how likely are you to suggest Chingu to a friend or colleague?",
+                        parseConfig: {
+                            "Not Likely": "Not Likely",
```
cherylli, line 98: For this maybe we need to use something like min / max for the keys
```js
parseConfig: {
  min: "Not Likely"
  max: "Extremely Likely" 
}
```

prisma/seed/forms/checkinform.ts
```diff
@@ -158,6 +167,10 @@ export const populateCheckinForm = async () => {
                             },
                         },
                         text: "{{Yes,No}}Did you deploy to Production at the end of this Sprint?",
+                        parseConfig: {
+                            Yes: "Yes",
```
cherylli, line 171: should we have `yes/no` (instead of "Yes/No") for the keys to be consistent with the naming?

prisma/seed/forms/checkinform.ts
timDeHof: Yes, I think that would make it easier to attach an icon or css settings to icon on the frontend.

cherylli (approved):
code looks good
db push and re-seed ✅
ran all tests ✅

Ajen07 (approved):

